Issue #128, opened 2014-11-12
problem import ._core -> No module named urllib.parse


ishankhare07:
File "rpc/websocket/**init**.py", line 21, in <module>
    from ._core import *
  File "rpc/websocket/_core.py", line 43, in <module>
    from six.moves.urllib.parse import urlparse
ImportError: No module named urllib.parse

(python2.7.5+)

sarietta:
I have the same issue. I have uninstalled six and re-installed to versions 1.6, 1.8, and 1.9 but the problem persists.

liris:
It works on my python 2.7.9 and six 1.9.

Does this work?

>  python -c "import six.moves.urllib.parse"

ishankhare07:
sent a [pull request] (https://github.com/liris/websocket-client/pull/148), it will be able to handle both an installed library on the module search path as well as a six.py file outside the package, i.e. 
``` shell
$ ls
websocket/
six.py
```

jayvdb:
@ishankhare07 , why cant you alter the module search path to include the directory where the six.py resides?

ishankhare07:
I don't understand why/how but now it works just out of the box. i.e. if my directory structure is :-
websocket/
six.py
myfile.py

and i do 
``` python
import websocket
```
it works, earlier it didn't. Also inside websocket/_core.py
``` python
import six
```
while the correct path to import six from websocket/_core.py should be
``` python
from ..websocket import six
```
am I missing something? or has the relative import behavior changed a bit since then (this issue dates to Nov last year)

jayvdb:
"from ..websocket import six" isnt correct; it is strongly discouraged.  "import six" is correct.

liris:
I won't merge #148 because this is the enviroment issue like module search path. We shouldn't include six module inside of our library.

Please tell the following result.

```bash
echo $PYTHON_PATH
python -c "import six; print(six.__file__)"
python -c "import six; print(six.__version__)"
python -c "import six.moves.urllib.parse"
```

engn33r:
As far as I can tell, this issue has not been a problem for the last few years and was likely a problem with an old version of six or a certain environment configuration. If you encounter this issue again with an updated setup and the latest code base, please share.

PR #148, opened 2015-02-14
fixing the issue of #128

ishankhare07:
providing a relative import
comaptible with python 2.7 and 3.4

### DIFF
--- a/websocket/_core.py
+++ b/websocket/_core.py
@@ -20,8 +20,15 @@ Copyright (C) 2010 Hiroki Ohtani(liris)
 """
 from __future__ import print_function
 
+try:
+	import six	#try to import an installed library
+except ImportError:
+	try:
+		from ..websocket import six	#try to import six.py at ../websocket
+	except ValueError:
+		print('Please install six')
+		exit()
 
-import six
 import socket
 
 try:


websocket/_core.py
```diff
@@ -20,8 +20,15 @@
 """
 from __future__ import print_function
 
+try:
+	import six	#try to import an installed library
+except ImportError:
```
jayvdb: People need to see this exception.

websocket/_core.py
```diff
@@ -20,8 +20,15 @@
...
+	try:
+		from ..websocket import six	#try to import six.py at ../websocket
+	except ValueError:
+		print('Please install six')
+		exit()
```
jayvdb: Never call exit() / sys.exit() in a library.

ishankhare07: so should i just leave it at the print line, because doing so would result in an error later on as soon as the name 'six' is used?

jayvdb:
I'd say the best approach now is to discuss your problem on the issue, as this code is adding package management to a package, and in the process degrades the package in an attempt to solve what looks like an environment problem that you and someone else are experiencing.  Lets try to figure out the environment problem first.

ishankhare07:
can we just leave the ImportError raised so that any importer would know if six is not installed plus we would also be able to use six as a file on the current directory

liris:
I wouldn't include six module into our library.

Issue #160, opened 2015-03-24
How to get response header after websocket connection established?


RicterZ:
I try to find a way to get response header, but the WebSocket class doesn't provide a method or attribute to get it.

liris:
Our library doesn't provide such a method.
Why do we need it?

RicterZ:
Sometimes other programs need the status code and the response headers in the context I think.
And in fact I met the problem so I have to fork it and change some code, it's very inconvenient.

PR #161, opened 2015-03-24
add status code and response headers for WebSocket class #160

RicterZ:
#160


liris:
Thanks for your patch but I've fixed with 3495bbf7e570fed0a90ae6db7427efbccfd869ae

RicterZ:
Thank you very much ;)

PR #171, opened 2015-04-13
Add new argument "host_only" in the options.

birdstar:
Currently, all "wss" or "https" protocol will add "443" at the end of the "Host" field. But for some reverse_proxy server, the server settings does not allow 443 port, but delegate the original host to a certain internal port. So it would better to provide a way to disable the default behavior.
Default value: host_only=False

How to use it?
In the ws.connect method:
ws.connect("wss://demo.net/",  cookie=CookieString, origin=Origin, host_only=True)

### DIFF
--- a/websocket/_handshake.py
+++ b/websocket/_handshake.py
@@ -67,7 +67,7 @@ def _get_handshake_headers(resource, host, port, options):
     headers.append("GET %s HTTP/1.1" % resource)
     headers.append("Upgrade: websocket")
     headers.append("Connection: Upgrade")
-    if port == 80:
+    if port == 80 or ("host_only" in options and options["host_only"] == True):
         hostport = host
     else:
         hostport = "%s:%d" % (host, port)


websocket/_handshake.py
```diff
@@ -67,7 +67,7 @@ def _get_handshake_headers(resource, host, port, options):
     headers.append("GET %s HTTP/1.1" % resource)
     headers.append("Upgrade: websocket")
     headers.append("Connection: Upgrade")
-    if port == 80:
+    if port == 80 or ("host_only" in options and options["host_only"] == True):
```
liris: This would be better...

```python
if port in (80, 443):
```

liris:
>  for some reverse_proxy server, the server settings does not allow 443 port, but delegate the original host to a certain internal port. 

I didn't get this. It looks a mistake of the proxy configuration.

websocket/_handshake.py
birdstar: This is not my purpose. I would like to give user a way to disable adding 443 behind "host".

birdstar:
In my current work, our websocket server is behind the nginx server with a port 9443. And the nginx server only listen 80 port. 
In current logic, your webstocket-client code is hard coded, if the protocol is "wss" and "https", it will add "443" after "host". So this does not work for my case, the nginx reject my connection.
I think this is not necessary always to add 443 in webstocket-client, or at least, it is better to provide user a way to disable the default behavior. It should be more flexible as an excellent 3rd party package.

liris:
Does it mean the nginx server is waiting HTTPS protocol with 80 port?
If the nginx server is waiting with 80 port without SSL connection, I think you should connect with  "ws" protocol to the nginx server.

liris:
You can udpate Host header like this...

```python
ws.connect("wss://demo.net/",  host="demo.net")
```

dimaqq:
I agree with @liris, HTTP spec states "A "host" without any trailing port information implies the default port for the service requested..."

In the case of `ws://` scheme that's port `80`, and `wss://` that's `433`.

Note, that ignoring `(80, 433)` is wrong, it should be dependent on the scheme/protocol.

PR #184, opened 2015-05-13
fix regression of f6af085e1a393a7ef8b7d194330b46e16ab217e2

sim6:
Fix duplicated output
```
$ wsdump.py ws://echo.websocket.org
Press Ctrl+C to quit
> hi!
> hi!
< hi!
> 
```
Note that I only typed one "hi!"


sim6:
Thanks!

PR #185, opened 2015-05-13
fix output of wsdump.py with python3

sim6:
Before this:
```
$ python3 wsdump.py ws://echo.websocket.org
Press Ctrl+C to quit
> hi!
< b'hi!'
> 
```

### DIFF
--- a/bin/wsdump.py
+++ b/bin/wsdump.py
@@ -116,6 +116,8 @@ def main():
         while True:
             opcode, data = recv()
             msg = None
+            if six.PY3 and opcode == websocket.ABNF.OPCODE_TEXT and isinstance(data, bytes):
+                data = str(data, "utf-8")
             if not args.verbose and opcode in OPCODE_DATA:
                 msg = "< %s" % data
             elif args.verbose:


bin/wsdump.py
```diff
@@ -116,6 +116,8 @@ def recv_ws():
         while True:
             opcode, data = recv()
             msg = None
+            if six.PY3 and type(data) is bytes:
```
liris: I think the following would be better.

```python
if six.PY3 and opcode == websocket.ABNF.OPCODE_TEXT and isinstance(data, bytes):
    data = str(data, "utf-8")
```

1. we need to convert data only if the data is TEXT type.
2. TEXT data is utf-8 encoded. "ENCODING" is terminal encoding.

sim6: Oh! Yes, I'm sorry I was wrong using ENCODING. About the TEXT type, you are right but if it is not a TEXT type is converted to string too. Anyway, I amended it with your suggestion.

sim6:
Thanks!

PR #214, opened 2015-10-18
Python 2.6 str.format does not support unindexed parameters

jayvdb:


jayvdb:
based on https://github.com/liris/websocket-client/issues/83 , I assume that Python 2.6 is "use at your own risk" support level.
If you'd prefer to actively de-support Python 2.6, a few documentation fixes would make that more clear ;-)

Issue #237, opened 2016-02-18
Wrong documentation + incorrect implementation for `ping_timeout`?


mkonecny:
In the code it says

> ping_timeout: timeout(second) **if the pong message is not received.**

But this doesn't appear to be true. The only thing `ping_timeout` is used for is to compare when the last PING was sent, and I don't see it related to PONG messages:

```
                if ping_timeout and self.last_ping_tm and time.time() - self.last_ping_tm > ping_timeout:
                    self.last_ping_tm = 0
                    raise WebSocketTimeoutException("ping timed out")
```

I see PONG messages only being used to call a user-defined callback:

```
                    elif op_code == ABNF.OPCODE_PONG:
                        self._callback(self.on_pong, frame.data)
```

Nikhil13x:
Hello.. is this issue addressed?

mkonecny:
Addressed in #241

PR #241, opened 2016-03-21
implement ping/pong timeout

mkonecny:
should fix issue #237


engn33r:
@xiaolang90 see PR #611 which fixed what you described

PR #249, opened 2016-04-27
Clean-up code

allanlewis:
I know some people hate clean-ups like this: please forgive me. My reason for raising this is because I'd like to contribute some changes but the existing code doesn't meet various conventions, like PEP 8. I'd still be willing to contribute anyway, but merging this would help greatly.


allanlewis:
I've rebased to squash a fix for a (completely legitimate) test failure.

PR #285, opened 2016-12-06
fix opcode name error

willmcgugan:
This fixes a `NameError` in Python3.


nlevitt:
Also https://github.com/liris/websocket-client/issues/261#issuecomment-264479759

liris:
I believe this is fixed with  #286

PR #288, opened 2016-12-07
Fix incorrect variable reference in _app.py

hugoduncan:


hugoduncan:
I just noticed the other two fixes for this. Closing.

PR #293, opened 2017-01-04
Use poll rather than select.

dellis23:
select cannot select on file descriptor numbers > FD_SETSIZE (1024 on
most systems), limiting the number of connections that can be created.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -195,13 +195,16 @@ class WebSocketApp(object):
                 thread.setDaemon(True)
                 thread.start()
 
+            poller = select.poll()
+            READ_FLAGS = (select.POLLIN | select.POLLPRI |
+                          select.POLLHUP | select.POLLERR)
+            poller.register(self.sock.sock, READ_FLAGS)
             while self.sock.connected:
-                r, w, e = select.select(
-                    (self.sock.sock, ), (), (), ping_timeout)
+                events = poller.poll(ping_timeout)
                 if not self.keep_running:
                     break
 
-                if r:
+                if events:
                     op_code, frame = self.sock.recv_data_frame(True)
                     if op_code == ABNF.OPCODE_CLOSE:
                         close_frame = frame


dellis23:
One of my esteemed colleagues noted that this does break compatibility with Windows if it is indeed intended to run there.  If so, I can do something to fall back to select.

swordmaster2k:
Came across this issue today when trying to open 1000+ websockets for a load test it led to the exception:

_filedescriptor out of range in select()_

dellis23:
Yeah that's the exact use case I had -- load testing.

swordmaster2k:
Installing your branch through pip solved the issue I encountered for now. Hopefully this will be merged, just have to be careful on behavior on Windows platforms.

dellis23:
Awesome, glad to hear that.  Thanks for posting the error message.  Hopefully that'll help googlers in the meantime.

liris:
Does this work on Windows? I'm not sure because I don't have Windows.
Please let me know whether this works on Windows. If it works, I'll merge this

swordmaster2k:
I've tested this on Windows and it does appear to break it I get the following error: module 'select' has no attribute 'poll'

I'm still using this branch myself, a fall back should be added for Windows OSes to use select instead.

bubbleboy14:
Hey homies. Here's the solution: https://github.com/websocket-client/websocket-client/pull/375

liris:
now we support dispatcher, which customize event poller/selector.

```python
class PollDispatcher
   def __init__(self):
        self.poller = select.poll()
        READ_FLAGS = (select.POLLIN | select.POLLPRI |
                                    select.POLLHUP | select.POLLERR)
        self.poller.register(self.sock.sock, READ_FLAGS)

    def read(self, sock, callback):
        while self.app.sock.connected:
            events = poller.poll(ping_timeout)
            if events:
                callback()

app = WebsocketApp(..., dipacher=PollDispacher())

```

PR #305, opened 2017-03-09
Fix `insecure_pythons` list in setup.py

underyx:
Before this it generated:

    '2.4, 2.5, 2.62.7.0, 2.7.1, …'


underyx:
Oh, sorry, I thought I opened this on the old repo first, but #304 is actually here.

PR #328, opened 2017-06-22
renames key in sslopt dict, fixes #326

### LINKED ISSUE #326: Unkown kwarg 'ca_cert' when calling ssl wrap_socket()
Introduced with commit 3339bd9877a841dc2fbac26bce294c5536bc2886 websocket-client no longer works with standard Ubuntu 14.04.

Ubuntu 14.04LTS has python version 2.7.6 and in the "/usr/lib/python2.7/ssl.py" it expects the kwarg to be named "ca_certs", not "ca_cert"

aoberoi:
must use the same key in the `sslopt` dictionary when reading and writing to it.

as mentioned after the issue was closed, the issue still existed. i made this change locally and verified that the issue is fixed.

**NOTE:** the API for `connect()` (internally called by `websocket.create_connection()`) allows the user to specify `sslopts`. This dict param still uses the `ca_cert` key, so as to not introduce a breaking change to the API.

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -171,7 +171,7 @@ def _ssl_socket(sock, user_sslopt, hostname):
         certPath = os.path.join(
             os.path.dirname(__file__), "cacert.pem")
     if os.path.isfile(certPath) and user_sslopt.get('ca_cert', None) is None:
-        sslopt['ca_cert'] = certPath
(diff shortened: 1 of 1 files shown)


Roach:
This issue was raised in https://github.com/slackapi/python-slackclient/issues/189

Issue #4, opened 2011-06-19
URL parsing is broken


wolever:
With Python 2.5 and 2.6, url parsing is broken because urlparse doesn't know how to interpret the `ws` scheme.  Here's a fixed version that works correctly:

```
def _parse_url(url):
    """
    parse url and the result is tuple of 
    (hostname, port, resource path and the flag of secure mode)
    """
    if ":" not in url:
        raise ValueError("url is invalid")

    scheme, url = url.split(":", 1)
    url = url.rstrip("/")

    parsed = urlparse(url, scheme="http")
    if parsed.hostname:
        hostname = parsed.hostname
    else:
        raise ValueError("hostname is invalid")
    port = 0
    if parsed.port:
        port = parsed.port

    is_secure = False
    if scheme == "ws":
        if not port:
            port = 80
    elif scheme == "wss":
        is_secure = True
        if not port:
            port  = 443
    else:
        raise ValueError("scheme %s is invalid" % scheme)

    if parsed.path:
        resource = parsed.path
    else:
        resource = "/"

    return (hostname, port, resource, is_secure)
```

brandonmoser:
This issue was reintroduced in a recent version. Changing the Scheme back to HTTP when parsing the URL allows the library to work on 2.6.1 (I haven't tried 2.5.x).

PR #332, opened 2017-06-30
Fixing parsing issue in Python 2.6.1

brandonmoser:
(Re)Fixes issue #4

### DIFF
--- a/websocket/_url.py
+++ b/websocket/_url.py
@@ -42,7 +42,7 @@ def parse_url(url):
 
     scheme, url = url.split(":", 1)
 
-    parsed = urlparse(url, scheme="ws")
+    parsed = urlparse(url, scheme="http")
     if parsed.hostname:
         hostname = parsed.hostname
     else:


ztemac32:
~/g/f/w/bin (master|✔) $ python wsdump.py \
  wss://www.bitmex.com/realtime/websocket?accessToken=xnyXWIQL0sbrLC91ISDlFhuLUGQJNq05A...
Press Ctrl+C to quit

  < {"info":"Welcome to the BitMEX Realtime API.","version":"1.1.0",
     "timestamp":"2015-01-18T10:14:06.802Z","docs":"https://www.bitmex.com/app/wsAPI","heartbeatEnabled":false}

> "help"

  < {"info":"See https://www.bitmex.com/app/wsAPI and https://www.bitmex.com/explorer for more documentation.",...}

> {"op": "subscribe", "args": ["trade:XBTUSD","instrument:XBTUSD"]}

  < {"success":true,"subscribe":"trade:XBTUSD",
     "request":{"op":"subscribe","args":["trade:XBTUSD","instrument:XBTUSD"]}}
  < {"success":true,"subscribe":"instrument:XBTUSD",
     "request":{"op":"subscribe","args":["trade:XBTUSD","instrument:XBTUSD"]}}

  < {"table":"trade","action": "partial", ...}
  < {"table":"instrument","action": "partial", ...}

  < {"table":"instrument","action":"update","data":[{"symbol":"XBTUSD","fairPrice":380.52,
     "markPrice":380.52,"indicativeSettlePrice":380.52,"timestamp":"2016-01-15T22:33:15.000Z"}]}
  < {"table":"instrument","action":"update","data":[{"symbol":"XBTUSD","openValue":673520400,
     "timestamp":"2016-01-15T22:33:15.000Z"}]}

liris:
I don't have python 2.6.1 environment. The following code works fine on python 2.6.9

```python
import urlparse

parsed = urlparse.urlparse("ws://loclhost:8080/foo")
print parsed
parsed = urlparse.urlparse("//localhost:8080/foo",  scheme="ws" )
```

engn33r:
This PR is quite old, and it seems to be relevant only for Python 2 (now EOL), but I do not currently see an issue with it. My only hesitation is that it might be possible to remove the `scheme` parameter from the urllib.parse() call entirely, which would be my preference if there is no impact to functionality.

My thought process: Based on my understanding, [urllib.parse() in Python 3](https://docs.python.org/3/library/urllib.parse.html) does support the `ws` and `wss` schemes. However, this documentation for Python 2 does not indicate support for `ws` or `wss` schemes. The scheme value in urllib.parse() doesn't really matter, since `parsed.scheme`, the only value impacted by this PR, is never used. Instead, the scheme value used is currently parsed from the [url.split() line](https://github.com/websocket-client/websocket-client/blob/00aeb2b4ce341cbd4b0650e373f1404a5b02f7dc/websocket/_url.py#L49) before the urllib.parse() call.

engn33r:
Except for backported fixes, this is expected to be the last commit for Python 2 support.

PR #345, opened 2017-08-30
Resolve a proxy issue and a connection error

bpmckinnon:
I made these two changes to be able to run websocket-client in my environment.  I have HTTP_PROXY set but the websocket server I'm trying to use is local.  I also needed to catch the WSAECONNREFUSED error which was being thrown while attempting to connect to the IPv6 address of the host.  I believe this error is windows specific, so I used the try/except.

I tested it in Windows 10 with python 2.7.8

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -114,7 +114,11 @@ def _open_socket(addrinfo_list, sockopt, timeout):
             sock.connect(address)
         except socket.error as error:
             error.remote_ip = str(address[0])
-            if error.errno in (errno.ECONNREFUSED, ):
+            try:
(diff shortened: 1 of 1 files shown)


bpmckinnon:
errno.WSAECONNREFUSED is not defined on Linux systems so it would fail.

I used the try because everyone tells me try/except is "pythonic."  I'm a c++ developer, so it kills me inside to use exceptions for control flow... but when in Rome.

bpmckinnon:
Hello,
Just wanted to follow up on this pull request.  Is there anything else that I can do?
Brian!

PR #356, opened 2017-09-26
add lock to recv function

eendebakpt:
The `enable_multithread` option created a lock, but it is only used when sending messages. This PR adds to locking to receiving as well.

### DIFF
--- a/websocket/_core.py
+++ b/websocket/_core.py
@@ -94,8 +94,10 @@ class WebSocket(object):
 
         if enable_multithread:
             self.lock = threading.Lock()
+            self.readlock = threading.Lock()
         else:
             self.lock = NoLock()
+            self.readlock = NoLock()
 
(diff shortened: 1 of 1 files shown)


totaam:
This may be a stupid question as I haven't looked at the code, but is it really safe to use the same lock here? Can't we have threads for each?

eendebakpt:
I don't know any details about websockets, but in general you don't want reading and writing to occur at the same time. In any case it is safer then using no lock at all.

totaam:
I don't know why you avoid reading and writing simultaneously, we do it all the time with a thread for each op.
Using the same lock means that those threads will now end up being synchronized, and this could conceivably deadlock.

eendebakpt:
@totaam Ok, I updated the code so that two different locks are used now.

PR #367, opened 2017-11-22
Init logging properly

flipreverse:
Init logging properly. Otherwise a user won't see any messages. Instead, a user will see the following error 'No handlers could be found for logger'. 
This issue occurred when I wrote faulty code in my on_message handler. As a result, I got the above error.


liris:
you should set up logging property prior to import websocket by your app.

Issue #368, opened 2017-11-30
Empty String not correctly passed as Origin Header


Marenthyu:
When passing ``origin = ""`` to ``.connect()`` or ``WebSocketApp.run_forever()``  this is not correctly translated to the sent Origin Header; Instead the default value of ``"http://" + hostname`` is set.

I came across this as i was connecting to a Websocket that required origin to be non-existant or set to an empty String (or a specific other value, which i am using as a workaround for now, but makes the origin header not true as that is not the actual origin.)

https://github.com/websocket-client/websocket-client/blob/7e88a5cc458739f2ed1b1a5f39f28d64e4d9e842/websocket/_handshake.py#L93 should be adjusted and not test for ``and options["origin"]`` - unless there is a reson for it. ``is not None`` may be the more appropriate test here.

liris:
I merged #369

PR #369, opened 2017-11-30
Update _handshake.py

Marenthyu:
Untested, but unless there is a reason "" should be treated the same as None, this should work fine.

Also includes the same "fix" for host.

See #368


liris:
This causes to send empty Host header.

Marenthyu:
That is **exactly** the intended result. Even if it seems stupid for most use-cases, if the user wants to deliberately set an empty host and/or Origin, it should be possible.
Please reconsider this PR to make this possible; for reasons why, see #368 and referencing Issues.

Issue #372, opened 2017-12-09
Broken self.keep_running flag 'false' value


mihai-mocanu:
Seems that https://github.com/websocket-client/websocket-client/pull/365 broke the false value form `False` to `false`

```
File "/usr/local/lib/python3.5/dist-packages/websocket_client-0.45.0-py3.5.egg/websocket/_app.py", line 102, in __init__
    self.keep_running = false
NameError: name 'false' is not defined```

liris:
Dip #371

PR #373, opened 2017-12-09
Update _app.py

mihai-mocanu:
Fix False typo see: #372


liris:
Duplicated with #372

liris:
#371

PR #386, opened 2018-01-23
Fix for issue 384 - WebSocketApp does not poll for data correctly when using SSL

TristenHayfield:


TristenHayfield:
I just realized that my trick of setting self.rselect in WebSocket.connect on the fly would leak memory. I'm about to push a version that remedies this.

TristenHayfield:
Rectified the aforementioned issue.

PR #389, opened 2018-01-24
Simplified Dispatcher API

bubbleboy14:
Love the new run_forever(dispatcher=something) interface. I have one suggestion to make it simpler and more broadly applicable.

I removed explicit dispatching for two reasons: 

 - by default, this function is empty, and isn't even called
 - this allows users to easily drop in dispatchers, such as pure-Python [rel](https://github.com/bubbleboy14/registeredeventlistener) and the ever-popular C-based [pyevent](https://www.developerfusion.com/project/39066/pyevent/)

With this API, multi-headed applications are quite simple:

```
import rel, websocket

def on_message(ws, message):
	print ws.symbol, message

def spawn(symbol):
	ws = websocket.WebSocketApp("wss://api.gemini.com/v1/marketdata/%s"%(symbol,),
		on_message=on_message)
	ws.symbol = symbol
	ws.run_forever(dispatcher=rel)

for symbol in ["BTCUSD", "ETHUSD", "ETHBTC"]:
    spawn(symbol)

rel.signal(2, rel.abort) # Keyboard Interrupt
rel.dispatch()
```

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -244,7 +244,6 @@ class WebSocketApp(object):
                 return True
 
             dispatcher.read(self.sock.sock, read)
-            dispatcher.dispatch()
         except (Exception, KeyboardInterrupt, SystemExit) as e:
             self._callback(self.on_error, e)
             if isinstance(e, SystemExit):
@@ -260,10 +259,6 @@ class WebSocketApp(object):
                     (self.sock.sock, ), (), (), ping_timeout or 10) # Use a 10 second timeout to avoid to wait forever on close
                     if r:
                         callback()
-
-            def dispatch(_):
-                # do nothing
-                pass
 
         return Dispatcher()
 


liris:
looks nice. I'll merge.
But, this is not a dispatcher. I want to change the name. It is hard for me to name it. Do you have any idea?

liris:
```python
class Selector:
    def select(self, socket, callback):
        r,w = select.select(xxx)
        if r:
           callback()

app.run_foreever(selector=Selector())
```

bubbleboy14:
I like that idea except that it's a little imprecise, because something like rel or pyevent (I think "dispatcher" is a decent abstraction) uses a variety of kernel-specific techniques (such as epoll, poll, kqueue, and select) under the hood. So since technically it's not always using "select", I feel that "Dispatcher" is a somewhat better term. What do you think @liris?

bubbleboy14:
Also, because in the code example at the top, the "dispatcher" (rel) is the thing that actually dispatches (the line at the end, rel.dispatch() -- same terminology in similar libraries such as pyevent).

So it's the thing tasked with eventually dispatching the event loop, once all the machinery is set up for each client in run_forever().

I think this especially makes run_forever(dispatcher=whatever), and the default of None, make sense, because without some specified external dispatcher, run_forever() simply does that work (of dispatching) itself :)

bubbleboy14:
Hello all. I heard there was an issue with custom dispatcher (rel and pyevent) support, which is necessary for asynchronous (including multi-WebSocketApp) setups. It's related to the check_callback function, which is passed to read().

I can't do anything about other event engines (such as pyevent), but I pushed a new version of rel ([0.4.4](https://pypi.org/project/rel/)) that supports websocket-client's new interaction model. It just requires one additional line of code:

`rel.safe_read()`

You just need to call safe_read() at some point before creating your WebSocketApp - before the "spawn(symbol)" loop in the example code in the description, for instance.

LMK if you have any issues.

PR #403, opened 2018-02-22
Fix the issue that websocket status message may not present

kevinxw:
It is observed that in some websocket server implementation, the status message in the header may not be set.  In this case using this library will pop up a 'list index out of range' error.

Add a check before we extracting the status_message to avoid the issue.

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -242,7 +242,8 @@ def read_headers(sock):
 
             status_info = line.split(" ", 2)
             status = int(status_info[1])
-            status_message = status_info[2]
+            if len(status_info) > 2:
+                status_message = status_info[2]
         else:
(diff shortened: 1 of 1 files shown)


gabstopper:
I've also recently run into this issue where Tomcat 8 only send back the 'HTTP/1.1 101' in the status line, whereas Tomcat 7 would send back 'HTTP/1.1 101 Switching Protocols'. 

According to rfc 2616, section 6.1.1:

"The Status-Code is intended
   for use by automata and the Reason-Phrase is intended for the human
   user. The client is not required to examine or display the Reason-
   Phrase."

See e.g. https://tools.ietf.org/html/rfc2616#section-6.1.1 for more reference.

PR #407, opened 2018-03-01
Allow server to not send a status text

### LINKED ISSUE #406: Crash when server doesn't return a Reason-Phrase (status text) after HTTP/1.1 101
If the HTTP server doesn't return `'HTTP/1.1 101 Switching Protocols` as it's supposed to do, but instead just `HTTP/1.1 101`, websockets-client crashes with the following stacktrace:

```
Traceback (most recent call last):
  File "./test.py", line 7, in <module>
    ws.connect('wss://www.example.com/')
  File "/usr/local/lib/python3.6/site-packages/websocket/_core.py", line 220, in connect
    self.handshake_response = handshake(self.sock, *addrs, **options)
  File "/usr/local/lib/python3.6/site-packages/websocket/_handshake.py", line 69, in handshake
    status, resp = _get_resp_headers(sock)
  File "/usr/local/lib/python3.6/site-packages/websocket/_handshake.py", line 133, in _get_resp_headers
    status, resp_headers, status_message = read_headers(sock)
  File "/usr/local/lib/python3.6/site-packages/websocket/_http.py", line 245, in read_headers
    status_message = status_info[2]
IndexError: list index out of range
```

I've fixed it locally, I'm just putting this here for the PR

gschizas:
Even though this doesn't conform to the HTTP protocol, the client should be resilient to the server's misbehavior

This resolves #406

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -242,7 +242,7 @@ def read_headers(sock):
 
             status_info = line.split(" ", 2)
             status = int(status_info[1])
-            status_message = status_info[2]
+            status_message = status_info[2] if len(status_info) > 2 else ''
         else:
             kv = line.split(":", 1)
             if len(kv) == 2:


liris:
dup with #403

PR #415, opened 2018-03-21
Load system default certificates if none are given

### LINKED ISSUE #413: Support secure SSL contexts without extra configuration
Python 2.7.9+ and 3.4+ [provide a function](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) to create a context that has hostname verification enabled and checks against the system certificate store. Using that to create the connection would make all usages of the lib secure by default, including wsdump.py.

minus7:
When not providing a CA file or path, load the system default certificate stores instead. This may fail (according to `ssl.create_default_context`)

Fixes #413. Compatibility with older Python versions untested. Should work on 3.4+, but I couldn't find info on Python 2.


minus7:
Removed the bundled cacert.pem to make this work as intended when the Python version supports loading default certs. It will not work without explicitly passing a cert bundle on other versions.

minus7:
FYI this should work on Python 2.7.9+,released 2014-12-10, in which `SSLContext.load_default_certs` was [backported](https://github.com/python/cpython/commit/daeb925cc88cc8fed2030166ade641de28edb396)

Issue #362, opened 2017-10-17
Socks5 proxy support


chiragrr:
I have a deployment scenario where I want to run WebSockets on top of a SOCK5 proxy 

[WS client ] --[Sock5 proxy]---[WS Server]

Right now I see that there is a support for HTTP proxy (with basic auth). How would we go ahead and add the support for SOCK5? 

I did a local build and by updating (hacking) def_open_socket I can use a SOCKS5 socket instead of regular sockets. 

> def _open_socket(addrinfo_list, sockopt, timeout):

Is this a change that we would be willing to take in upstream? Or should I branch?

Thanks,
Chirag

liris:
socket object can be passed to create_connection as parameter. if you set the SOCKS5 wrapper socket, you can communicate vi socks5 proxy.

sjtusmartboy:
is there any example of how we can use the SOCKS5 wrapper socket to be passed in as the parameter of create_connection? Thanks

Aliaxander:
> s there any example of how we can use the SOCKS5 wrapper socket to be passed in as the >parameter of create_connection? Thanks
+1
socks4,5,http proxy.

engn33r:
PR #417 added support for this. An example of using websocket-client with a proxy is in [the documentation of examples](https://websocket-client.readthedocs.io/en/latest/examples.html?highlight=proxy#connecting-through-a-proxy).

PR #417, opened 2018-03-30
socks5 via pysocks support

jhtitor:
After reading #362, I feel there might be little interest in this patch, but here goes nothing :)

This adds a new config option `proxy_type`, and when it's set to `socks4`, `socks5` or `socks5h`, pysocks gets invoked (so the dependency is soft). Although pysocks supports HTTP proxies too (and probably does it well), this patch chooses the original/internal websocket-client implementation for backwards-compatibility.

pysocks is kinda the de-facto standard for wrapping sockets into socks proxies, and having support for it built-in right into websocket-client, would be *enormously* helpful (to me, anyways). I do realize you can work around via `_create_connection`, but it gets really messy when you a) involve SSL and b) want to implement it to every websocket-based RPC you use.

Hopefully this is reasonable and/or helpful, please let me know if I can improve this patch somehow.

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -39,21 +39,71 @@ else:
 
 __all__ = ["proxy_info", "connect", "read_headers"]
 
+try:
+    import socks
+    ProxyConnectionError = socks.ProxyConnectionError
+    HAS_PYSOCKS = True
+except:
+    class ProxyConnectionError(BaseException):
+        pass
+    HAS_PYSOCKS = False
 
 class proxy_info(object):
 
     def __init__(self, **options):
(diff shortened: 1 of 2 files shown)


maxpy:
I use the socks5 proxy like this:
`

            ws = websocket.WebSocketApp(url,
                                        on_message = on_message,
                                        on_error = on_error,
                                        on_close = on_close)
            ws.on_open =on_open
            ws.run_forever(http_proxy_host="127.0.0.1",http_proxy_port=1080,proxy_type='socks5')
`
with,a shadowsocks proxy running on my computer and works well, but  ws can't connect to server

PR #425, opened 2018-05-02
fix: upgrade status message may missing

pentie:
a tomcat websocket impliment will send status line `HTTP/1.1 101 \r\n`, breaks our parsing.


liris:
dup with #403

PR #430, opened 2018-05-23
Change for inherit-ability.

Zwork101:
Right now, you can't really inherit the class. You would have to pass in the bound methods. That's kinda ugly, and should be unnecessary.


Zwork101:
#442 covers this much better

PR #444, opened 2018-06-19
Add option to disable sending the Origin header

hanazuki:
I'm writing an OpenStack console client using this library and want the library not to emit an `Origin` header because the console server expects the header not to be present when accessed from a non-browser client (This behavior accords to [RFC6455 &sect;1.3](https://tools.ietf.org/html/rfc6455#section-1.3) where non-browser WebSocket clients are not required to send the `Origin` header during a handshake). But there is currently no option to disable sending the header, and this patch implements it.

I chose `origin=False` option for the purpose because `origin=None` and `origin=''` are already used for other cases: sending the default origin (`"http://host:port"` where host and port are those of the WebSocket endpoint) and the emitting empty Origin header (which is illegal according to RFC6454 though), respectively.

### DIFF
--- a/websocket/_core.py
+++ b/websocket/_core.py
@@ -201,6 +201,7 @@ class WebSocket(object):
         options: "header" -> custom http header list or dict.
                  "cookie" -> cookie value.
                  "origin" -> custom origin url.
+                             Pass False to send no Origin header.
                  "host"   -> custom host header string.
                  "http_proxy_host" - http proxy host name.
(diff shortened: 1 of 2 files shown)


liris:
We expect options["origin"] should be string value. I think it is better to introduce option["suppress_origin"].

liris:
refs #HC459

hanazuki:
Thank you!

PR #468, opened 2018-08-22
Update _app.py

ihjmh:
if callback function belongs to an class ,line 333 will cause an exception, the "self" will missing
for example:
      myobj = MyClass()
      self.on_open = myobj.test()
this will nerver work

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -329,10 +329,7 @@ class WebSocketApp(object):
     def _callback(self, callback, *args):
         if callback:
             try:
-                if inspect.ismethod(callback):
-                    callback(*args)
-                else:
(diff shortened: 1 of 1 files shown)


MRSharff:
> for example:
myobj = MyClass()
self.on_open = myobj.test()
this will nerver work

I'm not sure if you mistyped this, but if you didn't, that is not working because you're setting `self.on_open` to the return value of `myobj.test()` (which I'm presuming does not return a class method), not the method `myobj.test`. Since the `self.on_open` is now set to None (again, assuming myobj.test() did not return a class method), the `inspect.ismethod(callback)` will return False.

I don't think this should have been merged @liris 
although, maybe I'm mistaken.

PR #441, opened 2018-06-17
Update supported Python versions (2.7, 3.4+)

hugovk:
Python 2.6 is EOL and no longer receiving security updates (or any updates) from the core Python team. 

It's also little used.

Here's the pip installs for websocket-client from PyPI for May 2018:
	
| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  68.09% |        583,338 |
| 3.6            |  16.99% |        145,548 |
| 3.5            |  12.25% |        104,991 |
| 3.4            |   2.41% |         20,624 |
| 2.6            |   0.12% |          1,012 |
| 3.7            |   0.11% |            918 |
| 3.3            |   0.02% |            192 |
| 3.8            |   0.01% |            105 |
| None           |   0.00% |             15 |
| 3.2            |   0.00% |              3 |
| Total          |         |        856,746 |

Source: `pypinfo --start-date 2018-05-01 --end-date 2018-05-31 --percent --markdown websocket_client pyversion`

Dropping it allows old code to be removed and other code to be updated for modern Python.

Here's a passing build for this PR: https://travis-ci.org/hugovk/websocket-client/builds/417574198

PR #474, opened 2018-08-29
SSL socket handling fix

juhhov:
Fix SSL socket read + write according to:
https://bugs.python.org/issue21430
https://docs.python.org/2/library/ssl.html#notes-on-non-blocking-sockets

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import sys
 from setuptools import setup
 import pkg_resources
 
-VERSION = "0.51.0-1"
+VERSION = "0.51.0"
 NAME = "websocket_client"
 
 install_requires = ["six"]
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import sys
 from setuptools import setup
 import pkg_resources
 
-VERSION = "0.51.0"
+VERSION = "0.51.0-1"
 NAME = "websocket_client"
 
 install_requires = ["six"]
--- a/websocket/_socket.py
+++ b/websocket/_socket.py
@@ -19,6 +19,8 @@ Copyright (C) 2010 Hiroki Ohtani(liris)
     Boston, MA 02110-1335  USA
 
 """
+import errno
+import select
 import socket
 
 import six
@@ -77,8 +79,24 @@ def recv(sock, bufsize):
     if not sock:
         raise WebSocketConnectionClosedException("socket is already closed.")
 
+    def _recv():
+        try:
+            return sock.recv(bufsize)
+        except SSLWantReadError:
+            pass
+        except socket.error as exc:
+            error_code = extract_error_code(exc)
+            if error_code is None:
+                raise
+            if error_code != errno.EAGAIN or error_code != errno.EWOULDBLOCK:
+                raise
+
+        r, w, e = select.select((sock, ), (), (), sock.gettimeout())
+        if r:
+            return sock.recv(bufsize)
+
     try:
-        bytes_ = sock.recv(bufsize)
+        bytes_ = _recv()
     except socket.timeout as e:
         message = extract_err_message(e)
         raise WebSocketTimeoutException(message)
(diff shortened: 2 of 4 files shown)


liris:
Let me confirm this change would work on Python 2.6?
I'm afraid this breaks 2.6.

jupe:
is python 2.6 support still needed? maybe CI should test against python 2.6 also to avoid unexpected breaks 🙂

jupe:
Okay, I saw discussion #441 - any plan to continue with it?  Did you try with py@2.6?

jacopomaroli:
I just saw this PR after opening mine here https://github.com/websocket-client/websocket-client/pull/511
The implementation proposed in this one is not compliant to the documentation provided because is missing the "while True" loop.
I can integrate the changes I made in this one and I can run some tests in python 2.6

juhhov:
Yes, it might be possible that recv fails many times, maybe the while loop is needed. Your implementation blocks indefinitely in select.select, https://docs.python.org/2/library/select.html#select.select. I believe that is not wanted behavior.

We are using own fork of websocket-client, but of course it would be better to be able to use the official one. Any help to get this merged is appreciated.

jacopomaroli:
how about a max_socket_retry for the loop and a max_socket_select_timeout configurable parameters set to some reasonable default? e.g. 5 retries and 3 seconds select timeout for both the recv and send methods.
If agreed I can proceed merging the changes in #511 to this PR

juhhov:
Are any of those needed?
`select.select` will timeout if no new data arrives (when timeout parameter is given) and maybe we can just use the same timeout that is assigned to socket for select.

juhhov:
Ok. Should consider also blocking sockets. Its timeout should be respected, at least to some extent.
How about you merge your changes and we'll continue from there?

jacopomaroli:
A blocking socket timeout would be respected when calling sock.recv. The select timeout is something different. Even if you set it to be the same value of the socket timeout as you suggest, you won't be guaranteed that the overall recv execution (and I mean the websocket-client library implementation in _socket.py which happens to be the same name of sock.recv) would not last more than the socket timeout value in case of a blocking socket: if a reading completely fails, it would be socket timeout + select timeout (which happens to be the same so in this case) you would get an actual overall timeout being the double of the socket timeout.
Also a blocking socket shouldn't throw a EAGAIN unless SO_RCVTIMEO has been enabled, and I couldn't find any example of a blocking socket throwing a EWOULDBLOCK (which makes sense).
In this case you can anyway disable the retries.

I updated #511 with what we discussed above

liris:
I cannot test this PR. Do we have some problem with this PR? Otherwise, I can merge this.

juhhov:
@jacopomaroli yep, I am aware that socket recv timeout is not always respected in this PR neither.
@liris we have been using this branch since August with no problems.

jacopomaroli:
Feel free to merge this and I can update my branch with these changes.
We can continue the discussion from there on #511

jupe:
any chance to get release with this soon?

liris:
I merged #474 and just relesased.
Do we have any other issues?

2019/02/23 15:55、Jussi Vatjus-Anttila <notifications@github.com>のメール:

trygveaa:
This breaks non-blocking sockets for me. What happens is that the socket throws SSLWantReadError, which is caught by websocket-client. Then select is called, but since the socket is non-blocking, timeout is 0, so it very likely still has no data. That means receive won't be called again and `_recv` will return None. This causes websocket-client to close the connection, which shouldn't happen.

I don't think a retry loop, which @jacopomaroli mentions and implements in #511, is an appropriate solution. That would mean that websocket-client would block the recv call from the application, which defeats the purpose of setting it to non-blocking.

I think the correct solution would be to just rethrow SSLWantReadError, so the application can handle it like we could before. When the socket is non-blocking, websocket-client shouldn't block either. Therefore, the application has to handle retries itself, which it can do by handling SSLWantReadError.

trygveaa:
@juhhov: Can you explain what issue you had, which this PR is solving?

As far as I can tell, you are using non-blocking sockets, but that would mean that `sock.gettimeout()` is `0`. In that case, `select.select` would return immediately, and therefore the socket is most likely still not ready. That means that `_recv` doesn't return anything, which causes the socket to disconnect. Is this not what's happening for you?

juhhov:
@trygveaa the application (SocketIO-client) that we are using this with has timeout.

I would prefer that this situation would already be handled by python's socket module. This is not the type of issue that each application should need to handle.

Anyway, two possible options to solve your problem I quickly came up with:
1. Throw SSLWantReadError and do not `select.select` if the socket timeout is 0
2. Default to some timeout with `select.select` if the socket timeout is 0

trygveaa:
> @trygveaa the application (SocketIO-client) that we are using this with has timeout.

Ah, okay. I got the impression that it didn't, but then I was wrong.

> I would prefer that this situation would already be handled by python's socket module. This is not the type of issue that each application should need to handle.

For blocking sockets, that makes sense. However, for non-blocking sockets, the applications has to handle it themselves, because websocket-client shouldn't block.

> Throw SSLWantReadError and do not select.select if the socket timeout is 0

Great, I'll make a PR for this.

trygveaa:
I submitted PR #536 as a fix.

Issue #479, opened 2018-09-13
WebSocketApp's on_close never emits status code or reason


aarondav:
The logic in websocket/_app.py is buggy -- we will never call on_close with the status code or reason from the close frame.

The bug was introduced in https://github.com/websocket-client/websocket-client/commit/8f6ef9a2a7e24b4f90e0a6cafe9d7ca3f1dc3047. Note that `close_frame` is defined as a method-scoped local variable in `def run_forever()`:

https://github.com/websocket-client/websocket-client/blob/efcf31b30122e05701fb3f696b029baf8faa6d8e/websocket/_app.py#L215

However, it is set in the inner method `def read()`, introduced in the commit mentioned above. Python scoping logic is such that setting `close_frame = ...` does not find and set the outer variable in `def run_forever()`, but rather creates a new variable in `def run()`. Therefore, `def teardown()` never observes the close_frame being non-None.

Experimentally, I observe this. My close_frame is not None when observing the close frame in read(), but teardown() doesn't see it, and therefore does not propagate it to the on_close handler.

engn33r:
This issue appears to have been resolved by PR #480 but the issue wasn't closed out

PR #480, opened 2018-09-13
[#479] Fix WebSocketApp on_close handling to support statusCode and reason

aarondav:
Issue described in #479. I have tested this fix manually on my own application, but could not find a good place to put a unit test for this behavior. The echo server does not seem to close connections of its own accord, save for timeouts.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -212,12 +212,16 @@ class WebSocketApp(object):
         if self.sock:
             raise WebSocketException("socket is already opened")
         thread = None
-        close_frame = None
         self.keep_running = True
         self.last_ping_tm = 0
         self.last_pong_tm = 0
 
-        def teardown():
+        def teardown(close_frame=None):
+            """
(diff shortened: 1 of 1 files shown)


elderlabs:
@engn33r so what is it I need to pass to the wsApp in order to enable/catch status codes and reason? I can see `teardown` and `close_frame`, but there's no clear documentation or definition in source to suggest what needs to be passed in order to receive these. I thought it might be in `enableTrace`, as the docs suggest it, but that's not the case here, specifically as I'd like to capture them in my application. Thanks.

engn33r:
@elderlabs Actually I think a new issue might be a better place for this, because I don't (currently) have a quick easy answer for this. If you could include a rough sketch of your current code and some pseudocode of what you are aiming for in the new issue, that will help me understand what you want to do. My guess is one of us will need to reverse engineer the source code to figure out where the close status info gets passed depending on how you want to use it.

Issue #257, opened 2016-05-20
auto redirect


alanzongsianshen:
can this websocket client library handle the case of 30X redirect automatically like the "requests" ?
[Requests: Redirection and History](http://docs.python-requests.org/en/latest/user/quickstart/#redirection-and-history)

liris:
This doesn't support redirection.
Why do we need this?

alanzongsianshen:
in this library, when it occurred a 30x return code, it threw except directly; however, all i want is a full response so that i can handle the redirection by myself. it all work out, when i altered the part of throwing exception.

liris:
If you provide the patch for redirection, I'll take it.

alanzongsianshen:
it is not a handle of redirection, it just passes the responses to client, not just throw exception during handshake so that client can have the new url and reconnect. the work i've done just more like a workaround by bypassing the _validate part and return status and response,  if the return code(status) is 30x.

partojkander:
@alanzongsianshen - There is now redirect support available (after merge of #491).

However, it doesn't return the response to the caller, but instead it automatically follows the redirects.

alanzongsianshen:
> @alanzongsianshen - There is now redirect support available (after merge of #491).
> 
> However, it doesn't return the response to the caller, but instead it automatically follows the redirects.
even better, Thx!

PR #491, opened 2018-10-11
Handle redirects in handshake

partojkander:
Support HTTP redirects as defined in RFC 6455.


partojkander:
This addresses the issue #257.

I have run the unittests, also with TEST_WITH_INTERNET enabled.

partojkander:
@totaam - fixed

PR #492, opened 2018-10-12
Status return for WebSocketApp.run_forever()

Mottl:
`WebSocketApp.run_forever()` returns `True` on any exception and `False` on `KeyboardInterrupt`.  
The purpose of this commit is to give a straightforward way to rerun `WebSocketApp` if it died unexpectedly:

```python
while True:
    ws = websocket.WebSocketApp(...)
    if not ws.run_forever():
        print("Bye")
        break
    else:
        sleep(5)
        print("Reconnecting...")
```

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -199,6 +199,11 @@ class WebSocketApp(object):
         origin: update origin header.
         dispatcher: customize reading data from socket.
         supress_origin: suppress outputting origin header.
+
+        Returns
+        -------
+        False if caught KeyboardInterrupt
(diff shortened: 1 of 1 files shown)


mtrojnar:
Great work!  This is extremely useful.
Shouldn't the example code be:
```python
ws = websocket.WebSocketApp(...)
while ws.run_forever():
    sleep(5)
    print("Reconnecting...")
print("Bye")
```
? There are no I/O operations in `websocket.WebSocketApp(...)`...

Mottl:
Yes, sure!

PR #502, opened 2018-11-07
fix callback

Lu-Yi-Hsun:
when my iqoptionapi use last version websocket-client get this error
```
error from callback <bound method WebsocketClient.on_message of <iqoptionapi.ws.client.WebsocketClient object at 0x7f174fdb5e48>>: on_message() missing 1 required positional argument: 'message'
  File "/usr/lib/python3.7/site-packages/websocket/_app.py", line 343, in _callback
   callback(*args)
```
change the code 
callback(*args)->callback(self, *args)
can fix my problem

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -340,7 +340,7 @@ class WebSocketApp(object):
         if callback:
             try:
                 if inspect.ismethod(callback):
-                    callback(*args)
+                    callback(self, *args)
                 else:
                     callback(self, *args)
 


engn33r:
Follow-up comment: both the "if" and "else" branch result in the same `callback(self, *args)` call. This is how the code was before PR #442, which broke other functions and was partially reverted. I properly reverted this portion of the code with commit 3112b7d

PR #510, opened 2018-12-09
adding ability to call instance method as callback

ofirule:
If I understand correctly, when someone inherits from WebSocketApp, and uses an instance method as a callback there is no reason to pass 'self' again to the callback.
but if someone uses an instance method of some other class the callback will fail with error:
on_open() missing 1 required positional argument: 'ws'
or similar

My fix allows passing 'self' only once when callback is an instance method of WebSocketApp, but won't fail when the callback is an instance of another class

I saw it was already part of the code once :
https://github.com/websocket-client/websocket-client/commit/ae589495b75162703ab4fe1f8712f50d22b12d4e

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -339,7 +339,7 @@ class WebSocketApp(object):
     def _callback(self, callback, *args):
         if callback:
             try:
-                if inspect.ismethod(callback):
+                if inspect.ismethod(callback) and isinstance(callback.__self__, WebSocketApp):
                     callback(*args)
                 else:
                     callback(self, *args)


jhtitor:
This should be merged ASAP, latest websocket-client release is breaking compatibility with older code without this fix.

engn33r:
This is a duplicate of PR #462. However, this callback branch should have been removed in the process of reverting PR #442. However, the revert was not done cleanly, so commit 3112b7d finally removes this callback branch and brings the callback back to where it was before PR #442. Due to commit 3112b7d, this PR is not longer relevant. If you would like to see or comment on the full history on this topic, please see [this comment](https://github.com/websocket-client/websocket-client/commit/3112b7d75b1e5d65cb8fdfca7801606649044ed1#commitcomment-47124551) on commit 3112b7d.

PR #511, opened 2018-12-11
Nonblocking websocket recv

jacopomaroli:
I'm experiencing quite a few issues because of this, hence the PR :)

When using non-blocking websockets the OS might throw EAGAIN or EWOULDBLOCK.
This is not really an error but rather a "please come back and retry later, buffer is still empty"
According to this https://stackoverflow.com/questions/11647046/non-blocking-socket-error-is-always
the function which reads from a non-blocking socket buffer should be handled as per this PR.

### DIFF
--- a/websocket/_socket.py
+++ b/websocket/_socket.py
@@ -56,6 +56,8 @@ class sock_opt(object):
         self.sockopt = sockopt
         self.sslopt = sslopt
         self.timeout = None
+        self.max_socket_retries = 5
+        self.max_socket_select_timeout = 3
 
 
 def setdefaulttimeout(timeout):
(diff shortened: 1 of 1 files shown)


websocket/_socket.py
```diff
@@ -73,21 +77,65 @@ def getdefaulttimeout():
     return _default_timeout
 
 
+def setmaxsocketretries(retries):
+    """
+    Set the maximum socket read/writes attempts.
+
+    retries: default socket read/writes attempts.
+    """
+    self.max_socket_retries = retries
```
liris: Where does this "self" come from? this method is not a class method.

trygveaa:
> When using non-blocking websockets the OS might throw EAGAIN or EWOULDBLOCK.

This is not something which should be handled by websocket-client, but something you have to handle in your application when using non-blocking sockets. When handling it in websocket-client like you do in this PR, it means that websocket-client will block the application, which defeats the whole purpose of using non-blocking sockets. I.e. this will break any application which relies on the socket being non-blocking.

engn33r:
Based on the above comments and where I think the line should be drawn between websocket-client library responsibilities and user application implementation responsibilities, I don't currently think code serving this purpose belongs in this library. If a user wishes to use non-blocking sockets, a loop, try statement, or other mechanism can be used to check for data without triggering an exception if there is no data received in the socket. If someone can provide at least 2 or more examples of other libraries that integrate this type of solution in the library, rather than in the user implementation, I would be open to changing my mind on this topic.

For anyone looking at this in the future, PR #474 has additional discussion that may be relevant to the issues in this library at the time when this PR was written.

Issue #471, opened 2018-08-25
on_open() missing 1 required positional argument: 'ws'


bitfag:
Please look at https://github.com/bitshares/python-bitshares/issues/136

The error caused by this change from `git diff v0.48.0 -- websocket/_app.py`:

```diff
     def _callback(self, callback, *args):
         if callback:
             try:
-                callback(self, *args)
+                if inspect.ismethod(callback):
+                    callback(*args)
+                else:
+                    callback(self, *args)
```

Probably the first callback should also receive `self`

MRSharff:
Seems like this was done so that you can use a method of an object that is not a WebsocketApp as a callback, as well as a function.

Imagine this scenario:

```
class SomeObject:
    label='this is some object'
    print_label(self):
        print(self.label)

someobj = SomeObject()
ws = websocket.WebsocketApp('ws://echo.websocket.org/', on_open=someobj.print_label)
```

when `_callback` calls `on_open`, `someobj` will pass itself into the `print_label` method, this will be *args in the _callback. If `self` is also on the first `callback` as you suggest, `someobj.print_label` will receive the `ws` object as `self` instead, and since the `ws` object doesn't have a `label` field, it will raise an `AttributeError`

bitfag:
@MRSharff though I see in e38d0e5611481dd4924004146a384867daaaaa4b this behavior was reverted to the previous version. Is it a final or it will be refactored again?

tonivega:
Workaround for the #futureme and maybe you too.

One might use this workaround for using a method from a class with WebSocketApp for "ws_open="

It works by using a closure around the original method from a class and returning a "valid" function (not method) for the ws_open callback caller.


    def ws_open(self, ws):
        ...some code here....

    def retro_callback(self, method):
        def final_call(*args):
            return method(*args)
        return final_call

    def ws_thread(self, *args):
        url = "wss://xxx"

        ws = websocket.WebSocketApp(url, on_open=self.retro_callback(self.ws_open), on_message=self.ws_message, on_error=self.ws_on_error)
`

PR #514, opened 2018-12-19
Ensure callbacks are called with proper arguments.

jhtitor:
Latest related commit is efcf31b30122e05701fb3f696b029baf8faa6d8e.
Related issues are #471.

Problem: if callback is a bound method of an object, _app code **wrongly** assumes the object is a subclass of WebSocketApp, and does NOT pass the "ws" parameter, thus breaking compatibility with existing code.

This fix allows all 3 use cases: WebSocketApp subclasses, unrelated objects and unbound functions.

```python3
def unbound_func(ws):
    print("WS:", ws)

class SomeSub(WebSocketApp):
    def my_on_open(self):
        print("WS:", self)

class SomeObject:
    def my_on_open(self, ws):
        print("WS:", ws)
```

Without this fix (or by reverting the whole `ismethod` thing completely), only one or two of those will work.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -339,7 +339,8 @@ class WebSocketApp(object):
     def _callback(self, callback, *args):
         if callback:
             try:
-                if inspect.ismethod(callback):
+                if (inspect.ismethod(callback)
+                and isinstance(callback.__self__, WebSocketApp)):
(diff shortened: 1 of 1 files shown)


jhtitor:
This is a duplicate of #510, sorry about that.

PR #533, opened 2019-03-14
_handshake: hasattr checks on six before accessing the values

sfoerster:
I had to make these changes to get websocket-client working in a kivy app with python3 where six did not have the PY34 attribute.

### DIFF
--- a/websocket/_handshake.py
+++ b/websocket/_handshake.py
@@ -31,13 +31,13 @@ from ._http import *
 from ._logging import *
 from ._socket import *
 
-if six.PY3:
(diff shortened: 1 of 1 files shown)


liris:
Can't you install the latest version of six?

sfoerster:
Sure. Can't we all just run Python 3.6.7 and raise exceptions otherwise? These lines of code exist to handle different versions in different environments and do a great job. With these hasattr checks they run exactly as they're supposed to and don't change any other behavior at all. I'm just trying to help you out.

PR #536, opened 2019-03-20
Don't retry sockets with timeout set to 0

### LINKED ISSUE #535: Websocket closes connection when SSLWantReadError is thrown
Just thought I'd open a ticket regarding this https://github.com/websocket-client/websocket-client/pull/474#issuecomment-472348867. Wee-slack is affected by this and the issue is tracked in https://github.com/wee-slack/wee-slack/issues/679.

@trygveaa wrote:
> What happens is that the socket throws SSLWantReadError, which is caught by websocket-client. Then select is called, but since the socket is non-blocking, timeout is 0, so it very likely still has no data. That means receive won't be called again and _recv will return None. This causes websocket-client to close the connection, which shouldn't happen.

PR https://github.com/websocket-client/websocket-client/pull/511 seems to be related to this too.

trygveaa:
For non-blocking sockets with the timeout set to 0, the select call in
_recv and _send will return immediately without the socket being ready.
This causes _recv to return None, which makes recv close the connection.
Since a select with a timeout of 0 right after recv/send is pointless,
just skip _recv and _send when timeout is 0.

As for the handling of SSLWantReadError, SSLWantWriteError, EAGAIN and
EWOULDBLOCK in _recv/_send, websocket-client can't handle these without
blocking. Since applications which has timeout set to 0 don't want
websocket-client to block them, they have to handle those errors
themselves.

Fixes #535

### DIFF
--- a/websocket/_socket.py
+++ b/websocket/_socket.py
@@ -96,7 +96,10 @@ def recv(sock, bufsize):
             return sock.recv(bufsize)
 
     try:
-        bytes_ = _recv()
+        if sock.gettimeout() == 0:
+            bytes_ = sock.recv(bufsize)
+        else:
+            bytes_ = _recv()
     except socket.timeout as e:
         message = extract_err_message(e)
         raise WebSocketTimeoutException(message)
@@ -148,7 +151,10 @@ def send(sock, data):
             return sock.send(data)
 
     try:
-        return _send()
+        if sock.gettimeout() == 0:
+            return sock.send(data)
+        else:
+            return _send()
     except socket.timeout as e:
         message = extract_err_message(e)
         raise WebSocketTimeoutException(message)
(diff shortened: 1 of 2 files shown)


rbrook:
Is it possible that this change is related to a "104, connection reset by peer" error that has been repeatedly thrown since I upgraded to 0.55.0?

trygveaa:
@rbrook: This change definitely doesn't cause it, since it was released in 0.56.0. It fixes a similar issue though, not sure if it helps for you. Have you tried upgrading to 0.56.0?

rbrook:
I have but unfortunately I'm still getting this error. Any ideas?
Another thing I noticed is once 1 connection is slammed with 104, the rest of the websockets in the process also terminate with 104.

PR #539, opened 2019-03-29
v should be checked for emptry string before splitting it

jpshivakavi:
when name 'v' has empty string on splitting this empty string we get a list with empty string like ['']
This will make the next if statement "if not no_proxy:" false.
This is leading to a bug where a connection initiated from a localhost / 127.0.0.1 will not be considered as no_proxy type of connection.
Leading to an exception ValueError("scheme %s is invalid" % scheme) with the scheme value set to http from parse_url() function


jpshivakavi:
When will it be merged with the master branch

jpshivakavi:
Could some one look into this and merge the fix.
This is my first bug fix ever in any open source project.

PR #568, opened 2019-09-02
Update to use the selectors module with DefaultSelector

### LINKED ISSUE #561: filedescriptor out of range in select()
hi. what is this message "filedescriptor out of range in select()" I receive this error in on_error() function when I increase load on my server?

DaanVanVugt:
This works from python 3.4 onwards and provides the most efficient
implementation depending on the current platform.
https://docs.python.org/3/library/selectors.html#selectors.DefaultSelector

Would fix #561 

This is a naive implementation, it might be possible to keep the selectors around for a longer time to avoid rebuilding them on every listen.


liris:
Travis failed to build. Could you check it?

tmatth:
> Travis failed to build. Could you check it?

Travis is failing on the Python 2.7 job, which as @Exteris notes in the description, is expected as this will only work for Python >= 3.4

indranilp:
This is still not available in master. Can you please check this?

engn33r:
I will merge this soon, since the discussion in issue #660 shows that this project will soon focus on only supporting Python 3 without the need for backward compatibility for Python 2.

engn33r:
Thanks for the PR!

DaanVanVugt:
Thanks for merging!

PR #587, opened 2019-12-02
Fix getting 400 bad request with long proxy authorization string

huaphuoctruong:
Previously passing a long username of password to proxy authorization will cause response 400 Bad Request from proxy server. This commit fix the problem.


izikorgad:
+1

PR #594, opened 2019-12-26
Use `thread.is_alive` as `isAlive` is deprecated

itaisteinherz:


itaisteinherz:
@liris It would be great if you could take a look at this :)

mikeage:
Note that this has been available since Python 2.7, so it's fully backwards compatible

seratch:
@engn33r I know you have been busy for checking many issues now but could you check this one if you have a chance? It'd be greatly appreciated if we can have this change in the next release!

Currently, I have been seeing this warning every time I stop an app using this library. As already mentioned above, merging this is safe enough as it's compatible with any Python versions this project supports.

engn33r:
Thanks @itaisteinherz @mikeage @seratch! All inputs are appreciated as we get this project back in shape

Issue #318, opened 2017-04-13
Pong befor Ping?


Nao83:
Shouldn't
[https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L228](https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L228)

be reversed? The pong message is supposed to arrive after the ping, hence its epoch time should be greater.
I think it also needs
`or self.last_ping_tm - self.last_pong_tm < 0`
In case the pong message never arrives.

Nao83:
[Updated line](https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L285)
In the if statement: 

1. The second line `time.time() - self.last_ping_tm > ping_timeout` guarantees we are checking at some time after the last_ping_tm + ping_timehout
2. Third line: if we have received a pong after the last ping we should check that has not arrived too late, so it should be `self.last_pong_tm - self.last_ping_tm > ping_timeout` (the current `self.last_ping_tm - self.last_pong_tm` is negative in this case!)
3. Also if the pong has not arrived at all we should throw the exception so we need `or self.last_pong_tm - self.last_ping_tm < 0` (last pong has been received before last ping, in other words no pong has been received after the last ping)

In summary it should be
```
if ping_timeout and self.last_ping_tm \
    and time.time() - self.last_ping_tm > ping_timeout \
    and (self.last_pong_tm - self.last_ping_tm > ping_timeout
         or self.last_pong_tm - self.last_ping_tm < 0):
    raise WebSocketTimeoutException("ping/pong timed out")
```

Or, in more readable code
```
has_timeout_expired = time.time() - self.last_ping_tm > ping_timeout
is_a_pong_not_arrived_after_last_ping = self.last_pong_tm - self.last_ping_tm < 0
has_pong_arrived_too_late = self.last_pong_tm - self.last_ping_tm > ping_timeout

if (ping_timeout 
        and last_ping_tm 
        and has_timeout_expired 
        and (is_a_pong_not_arrived_after_last_ping or has_pong_arrived_too_late)):
    raise WebSocketTimeoutException("ping/pong timed out")
```
What currently happens is that we only raise WebSocketTimeoutException if we send a ping after last pong + pong timeout!

purplecity:
谢谢  u very much!  I agree you .  you can request a code  merge

[...]

PR #611, opened 2020-03-28
'ping_interval' should be less than 'ping_timeout'

lano2088:


engn33r:
Thanks for the PR! I just ran into this same issue myself

patrickxia:
(moved from a comment on the commit as that's just hard to find).

This seems to just break things (ping timeouts will not occur as the check here won't be true: https://github.com/websocket-client/websocket-client/blame/833a3df1f956e3eaf8e4dbcddd19b5abd4263958/websocket/_app.py#L340).

`ping_interval > ping_timeout` is the expected state of affairs if you don't want to have more than one ping outstanding at any given time. Enforcing `ping_interval < ping_timeout` makes little sense. Maybe your problem was with equality instead of reversing the sign? Consider ensuring `ping_interval >= ping_timeout`. See, e.g. https://stackoverflow.com/questions/59763847/in-python-websocket-client-why-should-the-ping-interval-be-greater-than-ping-ti for discussion.

engn33r:
Thanks for the comment @patrickxia. That's interesting, I was under the impression that there are circumstances where a ping (or the corresponding pong) might encounter a network problem and therefore not be received before the next ping is sent. If such circumstances can arise, then I would expect multiple pings should be attempted before the connection is closed due to timeout. This is the standard behavior of the Linux "ping" utility - it sends more than one ping by default. I would think relying on equality of the timeout and interval values may induce race condition depending on the network, where timeouts sometimes occur if the network is congested. It's possible that line 340 also needs modification - I didn't look hard enough for side effects of this change (my next main focus is the increase code coverage, so that CI will catch such issues). But I could easily be misunderstanding some aspect of the purpose of the interval vs. timeout values, so I'm hoping you can weigh in on what I'm missing. One area I lack clarity is if the timeout counter only starts counting AFTER the ping is sent - if so, then my understanding is flawed, since I assumed the timeout counter and ping counter started at the same point in time. And thanks for linking to the stackoverflow conversation - I did not consider the impact of multiple pongs arriving in sequence if `ping_interval < ping_timeout` when a network problem does occur. I will take a closer look at this later, and try to trace back where the time source is for both counters, but wanted to send a quick reply to get your feedback on the flaws in my current understanding.

engn33r:
@patrickxia thanks again for pointing this out, I did not do due diligence in examining the implementation logic on this previously. You are right - the `ping_interval > ping_timeout` requirement as things were initially appears to be necessary for the line you referenced, containing `has_timeout_expired = time.time() - self.last_ping_tm > ping_timeout`, to ever be true. For posterity, and in case you want to weigh in, I will explain my step-by-step thought process, followed by some observations about the current implementation's weaknesses:

For a timeout to occur, BOTH `has_timeout_expired` and either `has_pong_not_arrived_after_last_ping` or `has_pong_arrived_too_late` must be true. For `has_timeout_expired` to be true, `time.time() - self.last_ping_tm > ping_timeout` must be true at some point in time (based on Line 340 that you linked to). We know that the left side of this inequality, `time.time() - self.last_ping_tm`, is always less than or equal to the ping interval, because a ping is sent at this interval regardless of whether a pong was received or not. Therefore, we can substitute `ping_interval` for `time.time() - self.last_ping_tm`. So for `has_timeout_expired` to be true, we must have `ping_interval > ping_timeout`, and that doesn't work with the `ping_interval < ping_timeout` check that I merged. So I will revert this change in a day or two in case you or others have any final comments on this.

I would like to add (for other readers, @patrickxia probably already realizes this) that the reason this confused me is that in the abstract (without looking at any code), the ping interval and ping timeout should not have any relationship, either <, >, <=, or >=. Pings should be sent whenever a user wishes, and the timeout should be the maximum time delay between a ping being sent and the corresponding pong response being received. If this time delay is exceeded, then the timeout should occur. The only reason there is a relationship requirement between the interval and timeout in the current code is because of the word "corresponding". Instead of tracking "time difference between sending ping x and receiving pong x", or "ping x+1 and pong x+1" (which is difficult because each ping must be paired with a specific pong and multiple can be in transit simultaneously), the currently uses simplified logic where 3 checks are used to emulate this behavior:
1. Has the time since the last ping exceeded the ping timeout?
2. Did the last pong arrive before the most recent ping?
3. Is the difference between the ping sent time and pong received time greater than the ping timeout?

The first of the 3 checks above creates the relationship (or limitation) between ping timeout and ping interval, because it prevents the scenario of multiple pings being sent before the timeout is reached. In theory, this limitation shouldn't exist - a user should be able to send a ping every 10 seconds but set a 1 minute connection timeout (ping_interval < ping_timeout). It's possible that the way threading is handled makes alternative options infeasible or difficult, but I don't know at the moment. I may create a new issue on this topic in a few days to bookmark this a topic for future improvement. I will also add a link to #318, which developed the logic currently used for the timeout.

PR #605, opened 2020-02-19
Capitalize default connection header

sawdustofmind:

PR #616, opened 2020-05-03
Fix typo in connection upgrade header.

nicolasavru:
See https://tools.ietf.org/html/rfc2817#section-3. Typo was introduced in commit 0598ef1a7a9d72670bf592439f1899412e790989.


nicolasavru:
Nevermind, dup of #605.

PR #623, opened 2020-07-26
Add is_opened() function in WebSocketApp class

G-Lauz:
As user sometime we want to know if the WebSocket his still open with the WebSocketApp class, used in that way could be intuitive:
```Python
import websocket

ws = websocket.WebSocketApp("ws://foo.com")
ws.run_forever()
if ws.is_opened():
    do_something()
```

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -312,6 +312,9 @@ class WebSocketApp(object):
             teardown()
             return not isinstance(e, KeyboardInterrupt)
 
+    def is_opened(self):
+        return self.sock is not None
+
     def create_dispatcher(self, ping_timeout):
         timeout = ping_timeout or 10
         if self.sock.is_ssl():


engn33r:
Sorry, I tried testing this PR and it did not work as expected for me. I will close this PR for now, but if you would like to provide a working proof of concept, code edit, or other update, I would be happy to reopen the PR it this works properly. Here is the code I used to test this PR (a lightly modified example from the [project documentation](https://websocket-client.readthedocs.io/en/latest/examples.html#creating-your-first-websocket-connection)):

```python
import websocket

def on_message(wsapp, message):
    print(message)

wsapp = websocket.WebSocketApp("wss://api.bitfinex.com/ws/1", on_message=on_message)
wsapp.run_forever()

print("Testing123")
if wsapp.is_opened():
    print(">>>>>>>>>>>>OPEN")
```

I do not see "Testing123" or "OPEN" printed out. With the way that the threading is currently functioning in the library, no code is executed after the `run_forever()` function.

PR #627, opened 2020-07-31
Add support for permitting additional values in the headers

### LINKED ISSUE #626: Header validation bug (Connection: Upgrade, Keep-Alive)
Connections will fail if the server responds with header "Connection: Upgrade, Keep-Alive" or any such variation with "Invalid WebSocket Header" which is not true.

This is the same issue as documented here: https://github.com/jlynch630/Harmony.NET/issues/2#issuecomment-435062514

grintor:
This pull request resolves #626 by allowing additional values (such as "keep-alive") to exist in the headers.

### DIFF
--- a/websocket/_handshake.py
+++ b/websocket/_handshake.py
@@ -178,8 +178,8 @@ def _validate(headers, key, subprotocols):
         r = headers.get(k, None)
         if not r:
             return False, None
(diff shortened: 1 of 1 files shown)


engn33r:
Thanks for the PR! I will add a unit test for this case

Issue #631, opened 2020-08-23
Need documentation for using Unix Domain Sockets


Rahul-RB:
As title says, we need some documentation on how to connect to web socket servers hosted at Unix Domain Sockets (because as far as I can see, nowhere in the readme does it show that the `WebSocket` or `create_connection` take a socket object as input).

PR #632, opened 2020-08-23
Added documentation for using Unix Domain Sockets and made corresponding changes to wsdump.py

Rahul-RB:
Issue reference: #631 

Adds a new option to `wsdump.py` called `--unix-socket`. This basically connects to the unix domain socket based websocket server. "URL" is just a valid Unix path.

Eg usage:

```
$ ./wsdump.py /path/to/unix.socket --unix-socket
> hello
< hello
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -151,6 +151,21 @@ You can also write your own class for the connection, if you want to handle the 
     ws = create_connection("ws://echo.websocket.org/",
                             sockopt=((socket.IPPROTO_TCP, socket.TCP_NODELAY, 1),), class_=MyWebSocket)
 
+Using Unix Domain Sockets
+---------------------------
+You can also connect to a WebSocket server hosted on a unix domain socket. Just use the `socket` option to setup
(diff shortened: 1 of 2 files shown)


engn33r:
Thank you for the PR! What I've done in 360ceff is I've taken your documentation input and placed it into the newly created project documentation [at this URL](https://websocket-client.readthedocs.io/en/latest/examples.html#using-unix-domain-sockets), which I think is a better way to document this project than relying on the README. I have not included the modifications to wsdump.py at this time because I would like to keep wsdump.py as a simple script to try things without additional setup effort. If you can provide an example of using wsdump.py with a unix socket with minimal effort, I can reconsider, but I think a better approach would be to create a new script to demonstrate unix socket functionality in the /examples directory that contains code for both creating a local WebSockets server and running using the unix socket option. I will happily consider a PR of such an example.

Rahul-RB:
Ahh sure, I'll do that.

PR #645, opened 2020-11-18
fix #595

### LINKED ISSUE #595: create_connection hangs on v0.57.0(tried v0.56.0 and works fine)
tested by using samsung tv v2 api

abma:
fix invalid case of "upgrade"


engn33r:
Thanks, this was addressed in #655 and should be fixed now

abma:
if so, #595 should be closed :-)

PR #656, opened 2021-01-25
Update _app.py

iunary:
Fix callback


xloem:
@lunary how does this compare to #502 ?  also are you available to help maintain this code?

engn33r:
Thanks for the PR! As xloem helpfully pointed out, this is similar to #502, and since 502 is already merged, I'm going to consider this issue as already resolved until further edge cases indicate otherwise. If you think this PR is superior to #502, please explain why as you might have a better understanding of the issue.

Issue #614, opened 2020-04-09
TypeError: '<' not supported between instances of 'SimpleCookie' and 'SimpleCookie' error


TaylorChen:
My local program tried to connect a few different servers with the socket, but the local program occasionally had an error in the header, after debugging it was found to be in the special case, _cookiejar.py The function ` CookieJar.get(host)` threw back an exception。after further debugging it was found to be the following code execution exception。
`return "; ".join(filter(None, ["%s=%s" % (k, v.value) for cookie in filter(None, sorted(cookies)) for k, v in sorted(cookie.items())]))`
When this exception occurs, cookies The variable is an array with the following values：
`[<SimpleCookie: SameSite='Lax' __cfduid='df1d897a2bf34846f61b020f8fa6c22ed1586410295'>, <SimpleCookie: SameSite='Lax' __cfduid='deeca9d96b18bc5cc41c1edd92afed0a51586410295'>]`。

I'm not sure if this is a bug in the websocket library, or if it's a bug on the subscription socket server, is there any solution?

PR #662, opened 2021-02-09
Simplified cookie sorting to fix comparison problem between SimpleCookies

nielsdraaisma:
Simplified sorting of cookies by sorting by the final string representation, this fixes https://github.com/websocket-client/websocket-client/issues/614.

### DIFF
--- a/websocket/_cookiejar.py
+++ b/websocket/_cookiejar.py
@@ -48,5 +48,5 @@ class SimpleCookieJar(object):
             if host.endswith(domain) or host == domain[1:]:
                 cookies.append(self.jar.get(domain))
 
-        return "; ".join(filter(None, ["%s=%s" % (k, v.value) for cookie in filter(None, sorted(cookies)) for k, v in
-                                       sorted(cookie.items())]))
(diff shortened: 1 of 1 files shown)


engn33r:
Thanks @nielsdraaisma, this PR is on my radar. I'll look at this more closely soon, but if you have any examples that illustrate the fix that this PR provides or if you can add a bit more explanation around the problem with the existing code, it might help me get to this sooner.

nielsdraaisma:
Thanks for checking in @engn33r , i've struggled to find the except set of cookies that cause this error, but the error that ends up in our application logs contains the following:

```
 '<' not supported between instances of 'SimpleCookie' and 'SimpleCookie'
```
This PR simplifies the sorting logic and I have not seen this error since switching to the fixed version.

engn33r:
@nielsdraaisma Thanks for the PR. In order to improve the cookie unit tests to try and catch similar issues to this in the future, could you provide a (redacted) example of the cookies or cookie pattern that prompted this error? From what I see of 614, it looks like cookies with the SameSite attribute is one situation that is not currently tested by websocket/tests/test_cookiejar.py.

Issue #559, opened 2019-07-03
http_no_proxy not handled correctly


thincal:
Actually the `http_no_proxy` could be configured as the postfix, but the matching logic in this library is just full matching the host with no_proxy.

engn33r:
This should be fixed with PR #671

PR #671, opened 2021-03-18
Match no proxy on domain and *

Codeglitches:
There is no real specification for the `no_proxy` environment variable, but according to [the curl manual](https://curl.se/docs/manual.html)

> A comma-separated list of host names that shouldn't go through any proxy is set in (only an asterisk, * matches all hosts)
>
> NO_PROXY
>
>If the host name matches one of these strings, or the host is within the domain of one of these strings, transactions with that node will not be proxied. When a domain is used, it needs to start with a period. A user can specify that both www.example.com and foo.example.com should not use a proxy by setting NO_PROXY to .example.com. By including the full name you can exclude specific host names, so to make www.example.com not use a proxy but still have foo.example.com do it, set NO_PROXY to www.example.com.

So NO_PROXY (or lowercase no_proxy) should account for:
* ip addresses
* ip ranges
* hostnames
* domainnames (which start with a dot)
* a star to match all

websocket_client does support the ip address, ranges and hostnames, but not domainnames and the star. I believe the lack of support of domainnames is mentioned in issue #559

This pull requests adds matching on domainnames and the star
It also:
* Moves unittests for the _url module to a separate test module
* Add unittests for testing the `_is_no_proxy_host` method which actually does the check
* Add unittests for testing the `_is_address_in_network`
* Fixes `_is_address_in_network`

The pull request should still work with python 2.7 (although implementing _is_address_in_network is a lot easier in python3)


websocket/tests/test_url.py
```diff
@@ -0,0 +1,225 @@
...
+import sys
+import os
+
+from websocket._url import get_proxy_info, parse_url
+
+if sys.version_info[0] == 2 and sys.version_info[1] < 7:
+    import unittest2 as unittest
+else:
+    import unittest
+sys.path[0:0] = [""]
```
engn33r, line 33: Why must the first sys.path value be an empty string? Maybe this is a common Python practice and I'm unaware.

engn33r:
Thanks for the PR! I haven't tested all the proxy edge cases, so expanding the test coverage is a good step forward. LGTM, but I added one minor question before I merge.

Issue #660, opened 2021-02-04
Project upgrade initiative


engn33r:
Hello and welcome! :wave:

You may have noticed that this repository has been in hibernation for a while, but it is in the process of being revitalized! After all, this amazing library is one of the [top 100 most downloaded Python modules](https://hugovk.github.io/top-pypi-packages/). As the new project maintainer (as of Jan 2021), I'm working on a few upgrades for this project, and there are many ways that you can help out! Here's a few ideas, but this list will change over time as we continue making progress here :raised_hands::raised_hands:

- The [new documentation](https://websocket-client.readthedocs.io/en/latest/) needs more details! If you have code snippets that can be added to the [examples](https://websocket-client.readthedocs.io/en/latest/examples.html) section, or problems that you already overcame that can go in the [FAQ](https://websocket-client.readthedocs.io/en/latest/faq.html), your input would be appreciated! :tada:
- If you are familiar with past issues in this project, there's a lot of dust that needs cleaning up! Bonus points if you can add to the [new documentation](https://websocket-client.readthedocs.io/en/latest/) and reference the documentation to close an issue :star:
- The [autobahn test code](https://github.com/websocket-client/websocket-client/blob/bdfdf4d7c8f14bb751cabdaf5f4171d98cf88669/compliance/test_fuzzingclient.py) in this project, under the compliance directory, uses an old autobahn installation method. It would be great to update this so we can keep track of performance improvements as the code changes. :+1:
- Are you familiar with GitHub Actions or CI best practices? Any additional automated tests that can streamline the process of testing new code would make it easier to apply upgrades. :sunglasses:
- Are you a Python 3 user? You should be! Python 2 was EOL in January 2021, but this project was focused on supporting Python 2 in the past. The migration to Python 3 is a work in progress, as is the migration away from focusing on Python 2, so if you see improvements here, please submit a PR. :smile:

Any other contributions are welcome! If you have any questions, concerns, or suggestions, just create an issue here and share your thoughts.

lalondesteve:
How brutal would you like the python3 transition to be? Should we assume that python2 is unsupported from now on?

engn33r:
[...]

PR #673, opened 2021-03-24
Add shebang

sedrubal:
**1. What is the problem with the current code**

Installing it (e.g. on archlinux) places `bin/wsdump.py` under `/usr/bin/wsdump.py`. Running `wsdump.py` directly executes that script. As there is no shebang, the system will use a shell interpretor (I think) to execute it. This results in weird behavior (e.g. calling the program `import` many times, that create something like screenshots)...

2. How your changes make it better

Add a shebang to let the system know, that this must be executed with python.

3. Provide some example code that can allow someone else to recreate the problem with the current code and test your solution (if possible to recreate).

```bash
$ wsdump.py --help
```

### DIFF
--- a/bin/wsdump.py
+++ b/bin/wsdump.py
@@ -1,6 +1,4 @@
-"""
-
-"""
+#!/usr/bin/env python3
 
 """
 websocket - WebSocket client library for Python


engn33r:
I am going to let this sit here a while because it is specific to Python 3. The timeline and plan around Python 2 vs. 3 support is still being determined in issue #660

sedrubal:
Oh I didn't expect that you are still supporting python2. In this case, you should add `#!/usr/bin/env python` as shebang. Otherwise people can't use the cli tool at all, I fear.

engn33r:
The [documentation currently suggests](https://websocket-client.readthedocs.io/en/latest/getting_started.html) `python wsdump.py`, but if you update the shebang PR to just `python`, I will accept that for now

sedrubal:
For my system it would be `python $(which wsdump.py)` which is unnecessarily complicated. But however, I changed the pull request ;)

PureTryOut:
Note that the current shebang in this PR will not work on systems that don't have Python 2 installed besides Arch Linux. Arch Linux is currently the only distro (afaik) that has `/usr/bin/python` be Python 3. Every other distro still has that as Python 2, and if you don't have Python 2 installed this script will just fail.

I would really recommend `/usr/bin/python3` or `/usr/bin/env python3` as shebang. Scripts should never really use `python`, they should always explicitly specify what version they need.

engn33r:
@PureTryOut @sedrubal Thanks for the thought, I guess having both Python 2 and Python 3 installed (as I do) will become less common in the future, and unless an alias exists, calling `python` without a version number can cause problems. I am reverting to my initial opinion and will let this sit here a while because such a change in its current form is specific to the Python version and version support strategies are still being discussed in #660 (comments are welcome). We can go back to using `python3` in the Python 3 branch when the split happens. This item doesn't seem particularly pressing at the moment since I would guess everyone can find a suitable workaround to run `wsdump.py` in the meantime, even if the workaround is not ideal.

sedrubal:
I'd suggest that you consider merging the first of my commits.

|   | no shebang | `#!/usr/bin/env python3` | `#!/usr/bin/env python` | 
|---|---|---|---|
| execute `wsdump.py`, python2 is installed, `python` is `python2` | ❌ | ❌ | ✅ |
| execute `wsdump.py`, python3 is installed, `python` is `python3`| ❌ | ✅ | ✅ |
| execute `wsdump.py`, python3 is installed, `python` does not exist | ❌ | ✅ | ❌ |
| execute `wsdump.py`, python2 & python3 is installed, `python` is `python2`| ❌ | ✅ | ✅ (`python2` is used) |
| execute `wsdump.py`, python2 & python3 is installed, `python` is `python3`| ❌ | ✅ | ✅ (`python3` is used) |
| execute `python $(which wsdump.py)`, python2 is installed, `python` is `python2` | ✅ | ✅ | ✅ |
| execute `python $(which wsdump.py)`, python3 is installed, `python` is `python3`| ✅ | ✅ | ✅ |
| execute `python3 $(which wsdump.py)`, python3 is installed, `python` does not exist | ✅ | ✅ | ✅ |
| execute `python $(which wsdump.py)`, python2 & python3 is installed, `python` is `python2`| ✅ (`python2` is used) | ✅ (`python2` is used) | ✅ (`python2` is used) |
| execute `python $(which wsdump.py)`, python2 & python3 is installed, `python` is `python3`| ✅ (`python3` is used) | ✅ (`python3` is used) | ✅ (`python3` is used) |

As you can see, there is an improvement in some cases, no scenarios break compared to the current solution, python3 is preferred and it is still compatible to python2. In my opinion there is no reason not to merge it. (But imho there is also no reason to support python 2 after its EOL, too).

mgorny:
This regression has been introduced as part of 'reverting license migration' in 6eaed48d49ea6a1a792b152a477bf9026f0c29b4. At the very least, the previous shebang should be restored. Discussing what shebang should be there for 2 weeks doesn't change the fact that the current release is broken and runs Python as shell script which is annoyingly broken in the best case (like when `import` from IM is started) and harmful in the worst case.

engn33r:
Sorry for the delay and change of mind. 0.59.0 was just released which is the last release to support Python 2, so now we can focus on Python 3 without worrying about backward compatibility. @sedrubal could you revert this PR to your original python3 shebang? I'll merge quickly after this is updated. Alternatively, tips on how to merge a specific commit in a PR would also work.

sedrubal:
I rebased the branch and dropped the second commit. So you can go on 😉 Thank you!

PureTryOut:
> Alternatively, tips on how to merge a specific commit in a PR would also work.

@engn33r not required for this PR anymore but for future, you can merge specific commits with `git checkout master && git cherry-pick <commits you want to merge> && git push`. You'll have to close the PR manually but the picked commits will be in master.

engn33r:
@sedrubal Thanks again for the PR and the thorough python vs. python3 comparison. I apologize for the delay and confusion on my end. All set now!

@PureTryOut Good tip, thanks. I'll test this out on my own soon. There's always more to learn with git.

PR #690, opened 2021-05-12
Fix: ResourceWarning unclosed socket

rmkraeuchi:
Close socket if error occurs in open_socket. This can happen if websocket is not available. 

**Current Code:**
```
url = 'ws://10.1.1.11:8081/ws'   # any unreachable url is working
ws = websocket.WebSocket()
ws.connect(url=url, timeout=0.1)
```

**Received warning in console:** 
ResourceWarning: unclosed <socket.socket fd=636, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('0.0.0.0', 56100), raddr=('10.1.1.11', 8081)>

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -184,6 +184,8 @@ def _open_socket(addrinfo_list, sockopt, timeout):
                     err = error
                     continue
                 else:
+                    if sock:
+                        sock.close()
                     raise error
             else:
                 break


engn33r:
Thanks for the PR. Good catch, I haven't checked if there are other places in this project where resources are not cleanly handled in error or exception cases.

PR #698, opened 2021-05-31
Add sslopt parameter 'server_hostname'

wildraid:
At this moment the hostname from connection URL always goes to "server_hostname" of [ssl.SSLContext.wrap_socket](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket). It makes perfect sense for most of use cases, but it's not possible to override it.

However, in some rare cases it might be necessary to pass a custom "server_hostname" parameter for SSL connections. For example, if we want to open an SSL connection to specific IP address, but certificate is valid for hostname only.

```
import websocket

ws = websocket.create_connection(f'wss://xxx.xxx.xxx.xxx:8563', {'sslopt': {'server_hostname': 'my_domain.local'}})
```

Thank you.

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -248,6 +248,9 @@ def _ssl_socket(sock, user_sslopt, hostname):
             and user_sslopt.get('ca_cert_path', None) is None:
         sslopt['ca_cert_path'] = certPath
 
+    if sslopt.get('server_hostname', None):
+        hostname = sslopt['server_hostname']
+
     check_hostname = sslopt["cert_reqs"] != ssl.CERT_NONE and sslopt.pop(
         'check_hostname', True)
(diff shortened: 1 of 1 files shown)


engn33r:
This looks reasonable, as I have seen there are a number of other ssl library options not currently supported in the websocket-client code. Before I examine it closely, can you consent to the Apache 2.0 license change in issue #697? Instructions are at the top of that issue. To avoid complications, I want to avoid accepting any PRs during this license transition unless the contributor first consents to the license change.

wildraid:
@engn33r , of course, no problem, I did it.

It's good to see `websocket-client` being updated and maintained once again. Thank you for that! :)

engn33r:
LGTM, thanks for the PR

PR #704, opened 2021-06-07
Fix spaces

varfigstar:
There are shouldn't be spaces when define values for attributes

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,10 +85,10 @@ def on_open(ws):
 if __name__ == "__main__":
     websocket.enableTrace(True)
     ws = websocket.WebSocketApp("ws://echo.websocket.org/",
-                              on_open = on_open,
-                              on_message = on_message,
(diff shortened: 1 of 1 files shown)


engn33r:

Good catch, at some point I want to add CI coverage for all the code coverage examples in this project. The flake linting GitHub Action would catch items like this. If anyone knows of an easy way to do that, do share. I haven't looked at how to implement this idea yet.

varfigstar:
Hello! I accept lecense in issue. Did you hear about Python Black? In two click you can rework code of file with PEP8 standarts. But it can request some postformating.

engn33r:
I don't know if black or other PEP8 tools work for .md or .rst files, like README.md (this PR) or the readthedocs documentation in this project under /docs/source. If there is a way to tell any PEP8 tool to check these other file extensions, let me know.

PR #710, opened 2021-06-25
support proxy auth

aladdinding:
usually proxies need authentication

### DIFF
--- a/bin/wsdump.py
+++ b/bin/wsdump.py
@@ -29,7 +29,7 @@ import time
 import ssl
 import gzip
 import zlib
-from urllib.parse import urlparse
+from urllib.parse import urlparse, unquote
 
 import websocket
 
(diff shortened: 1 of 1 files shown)


engn33r:
This PR only modifies the bin/wsdump.py file, and does not provide proxy auth for websocket-client in other cases, such as when `import websocket` is used in another application. Could you provide a more general implementation of this? It will involve modifying more than one file, but would be more beneficial for other users.

nyuszika7h:
I see what you did here with `unquote`, but the commit description is quite confusing, as `websocket-client` already supports proxy authentication, it just had a bug regarding urlencoded credentials. I've basically implemented the same thing with `unquote` in #723, but in the main code, and I didn't touch `wsdump.py`. Not sure if this patch is still required, as I haven't used wsdump.

engn33r:
Thanks @nyuszika7h, I was confusing this issue with #723 and I forgot I already merged your fix. I think including this fix in the main code is enough for now to avoid over complicating the wsdump.py example.

Issue #717, opened 2021-07-12
Race condition with on_close and socket close in multithreaded setup


Pithikos:
This is a bit of an edge case but it can be really hard to debug (it took me many days to figure out).

Scenario:
1. Server sends websocket `CLOSE`.
2. Client sends websocket `CLOSE`.
3. Server closes TCP socket.
4. Client's socket ought to to close at this point. However there are cases where this doesn't happen (from application's perspective - e.g. in another thread).

The problem is in this code:
```python
self._callback(self.on_close, close_status_code, close_reason)
self.sock = None
```

Now imagine you have this callback
```python
def _on_close(self, ws, close_status_code, close_msg):
     print(f"Disconnected from server. Will reconnect in 1000 seconds")
     self.on_close(close_status_code, close_msg)
     self.retry()
```

The retry can be blocking for a very long time like in this example. Another thread in the code can try to `send()` data and it will not get any error since the `BrokenPipe` will only be raised on the 2nd attempt. And hence a lost packet. This is really hard to debug and can go unnoticed.

Fortunately `sock.connected` is showing as `False` as expected, but this is not enough imo, since it expects the user of the library to know to check that specific boolean before trying to send a message..

**Solution**

We should invert the lines. The callback should be called once all destruction has already happened (closing sockets, etc.) so it doesn't potentially block the actual destruction of objects.

```python
self.sock = None
self._callback(self.on_close, close_status_code, close_reason)
```

engn33r:
I can imagine the troubleshooting for this issue took a while - nice find! Could you share which websocket server you were using, in case the server has something to do with it? I would like to try and replicate the problem scenario, so if you have more client example code to recreate the issue, especially if you can recreate it with a public websocket URL (for example, grep for "ws://" and "wss://" in the websocket/tests directory of this project), this would help.

Pithikos:
@engn33r sure. I was using https://github.com/Kai-Bailey/WebSock but actually ran into some other issues with that (https://github.com/Kai-Bailey/WebSock/issues/13)

So currently I'm in the process of extending my own library (https://github.com/Pithikos/python-websocket-server) which I haven't touched for 3 years. There's a chance I might add some test cases similar to this. Happy to share the code if I do so

All this is actually part of another much bigger project where I use your client in production and my server for imitating a 3rd party server in testing.

PR #718, opened 2021-07-12
Fix race condition

Pithikos:
This should solve edge cases like mentioned here: https://github.com/websocket-client/websocket-client/issues/717


engn33r:

engn33r:
@Pithikos sorry for the delay here. I'm trying to prioritize the Apache 2.0 license change since that's been ongoing for several months, then I'll get around to merging this PR. I would like to try and replicate the issue if I can, just to see the difference this makes for myself.

PR #721, opened 2021-07-13
Updating example for send_frame

shopped:
When I tried to run the example for send_frame on a local server (https://github.com/websockets/ws), I got an "op_code invalid" error. I don't think you can start a frame sequence with OPCODE_CONT, and OPCODE_TEXT by default sets the "fin" to true, or 1.

Sending the first message with a fin flag of 0 signals that the message is fragmented, and prevents the server from throwing an error. Not sure if this is unique to the server implementation I used, if it is please let me know! Otherwise I hope this can help someone else.

### DIFF
--- a/websocket/_core.py
+++ b/websocket/_core.py
@@ -286,7 +286,7 @@ class WebSocket(object):
         Send the data frame.
 
         >>> ws = create_connection("ws://echo.websocket.org/")
-        >>> frame = ABNF.create_frame("Hello", ABNF.OPCODE_TEXT, 0)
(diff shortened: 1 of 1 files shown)


engn33r:
I think this example works fine as is. FIN can be set to either 0 or 1, depending on whether the message contains multiple frames, as the 2nd and 3rd lines containing `ABNF.create_frame()` indicate. The example shows 2 separate messages being sent: the first is a OPCODE_TEXT message and the second is made of 2 OPCODE_CONT frames (which requires the first to have FIN=0).

Let me know if I overlooked something. Maybe this is an issue with specific servers.

PR #723, opened 2021-07-15
Unquote username and password in proxy from environment

### LINKED ISSUE #722: Authenticated proxy fails if username/password contains urlencoded characters
When I try to use an authenticated proxy like this:

```sh
export http_proxy=http://USERNAME:PASSWORD@example.com:1234
export https_proxy=http://USERNAME:PASSWORD@example.com:1234
```

I get `407 Proxy Authentication Required` with `websocket-client`, while it works fine with `requests` etc.

nyuszika7h:
This fixes e.g. credentials containing a `@` escaped as `%40`.

Closes #722.

### DIFF
--- a/websocket/_url.py
+++ b/websocket/_url.py
@@ -26,7 +26,7 @@ import os
 import socket
 import struct
 
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 
 __all__ = ["parse_url", "get_proxy_info"]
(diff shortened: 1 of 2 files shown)


engn33r:

nyuszika7h:
> First, because this project is undergoing a license change, can you consent to the Apache 2.0 license change in issue #697? Instructions are at the top of that issue. To avoid complications, I want to avoid accepting any PRs during this license transition unless the contributor first consents to the license change.

Sure, done.

engn33r:
Thanks for the contribution, and for adding a test case for this!

PR #728, opened 2021-08-11
Add python-socks

fabaff:
Is now a requirement.

- https://github.com/websocket-client/websocket-client/blob/master/websocket/_http.py#L35-L37
- https://github.com/websocket-client/websocket-client/blob/master/websocket/tests/test_http.py#L30-L31

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,9 @@ import pkg_resources
 
 VERSION = "1.2.0"
 
-install_requires = []
+install_requires = [
+    "python-socks",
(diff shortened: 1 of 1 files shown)


engn33r:
I am closing this PR in favor of commit 4951de2 to avoid adding a mandatory dependency to this project. I made some CI changes that are currently causing CI to fail, but I think this alternative approach will allow python-socks to remain optional for users who do not use a SOCKS proxy.

PR #732, opened 2021-08-22
Added documentation for sslopt

CheeseCrustery:
Multiple functions accept an `sslopt`-dictionary when creating a connection, but so far, there was no documentation on what the dictionary actually does. As different keys of `sslopt` are used on different functions and attributes of the `SSLContext` class, this is not self-explanatory.


engn33r:
Thanks for the PR!

Issue #730, opened 2021-08-18
Does WebSocketApp.run_forever() respond to pings?


Julian-O:
The [documentation](https://websocket-client.readthedocs.io/en/latest/examples.html#post-connection-features) states:

> The `run_forever()` function will automatically send a “pong” when it receives a “ping”, per the specification.

Terrific. This is the functionality I would like for free.

But when I have tracing turned on, I can see the remote server sending me pings, and no evidence of a pong being returned. 

[To avoid confusion: I am talking about responding the pings initiated by the remote end. Not initiating pings to ensure the remote end is present.]

So, I hunted through the code, looking for where it responds, and I cannot see it. It calls-back `on_ping()`, if provided, but doesn't automatically pong. 

Am I missing something? I don't know whether the functionality is missing, or I am having a bad day reading code.

I would have expected that, if `on_ping()` is not provided by the user, a default callback function would be used that performed the send the pong, but if `on_ping()` is provided it becomes the responsibility of the end-client to respond, so that they may write custom code that verifies more of the stack is functioning.

(Testing with V1.2.1.)

Julian-O:
> I can see the remote server sending me pings, and no evidence of a pong being returned.

Well, upon further investigation, the pings being sent by the third-party server are using `OPCODE_TEXT`, rather than `OPCODE_PING`, so they should **not** be automatically responded to. My question still stands though.

engn33r:
Looks like you're right. I suspect [this line](https://github.com/websocket-client/websocket-client/blob/5f32b3c0cfb836c016ad2a5f6caeff2978a6a16f/websocket/_app.py#L338), currently line 338 of _app.py, is where the ping should get automatically handled. Unfortunately I don't think any on_ping() callback function exists by default, so it looks like either the documentation and/or the code needs an update. Feel free to submit a PR if you have a fix in mind, otherwise I'll get around to this later.

As you pointed out, a server using `OPCODE_TEXT` for a ping would not trigger the on_ping() callback function.

Julian-O:
I don't need this functionality at the moment, because the server I am talking to is non-conformant anyway, but I am considering what would be required to make the change anyway to give back tot he community.

Here is my mulling out aloud.

# Code changes

* Define a public `send_pong()` method that transmits the appropriate OpCode, using send to do the work. Include comment that it is not necessary to call this, unless an `on_ping` parameter is provided.
* Define a private `_on_ping()` callback method that calls `send_pong()`
[...]

PR #733, opened 2021-08-23
Ping/Pong behaviour explained in Example doc.

Julian-O:
As discussed in #730. 
Demoted Post-Connection features. Added context for Autobahn. Copy edit.
Corrected misstatements about auto-responding pongs (Applies to WebSocket and WebApp).
Removed example of calling pong() - not necessary.
Explained the websocketapp example including auto-pinging and checks for response.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -397,12 +397,17 @@ is for a AF_INET (IP address) socket.
                           socket = my_socket)
 
 
-Post-connection features
-==========================
+Post-connection Feature Summary
+-------------------------------
 
-You can see a summary of this library's supported WebSocket features by either
-running the autobahn test suite against this client, or by reviewing a recently
-run autobahn report, available as a .html file in the /compliance directory.
+`Autobahn|TestSuite <https://github.com/crossbario/autobahn-testsuite>`_ is an
+independent automated test suite to verify the compliance of WebSocket implementations.
+
+Running the test suite against this library will produce a summary report of the
+conformant features that have been implemented.
+
+A recently-run autobahn report (available as an .html file) is available in the
+/compliance directory.
 
 Ping/Pong Usage
 --------------------------------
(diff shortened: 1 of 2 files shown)


docs/source/examples.rst
```diff
@@ -426,7 +436,6 @@ For debugging, remember that it is helpful to enable :ref:`Debug and Logging Opt
 
 This example is best for a quick test where you want to check the effect of a
 ping, or where situations where you want to customize when the ping is sent.
-This type of connection does not automatically respond to a "ping" with a "pong".
```
engn33r, line 429: At first I was confused because we were discussing ping/pong in the WebSocketApp `run_forever()` case, but a quick test confirms that you are correct about this comment being inaccurate. Good call

engn33r:
Do you want to find a place to add a comment about the `recv_data_frame()` function being the source of the automatic pong response? That way we don't lose that valuable insight from this effort!

docs/source/examples.rst
```diff
@@ -437,15 +446,18 @@ This type of connection does not automatically respond to a "ping" with a "pong"
...
-If a server needs a regular ping to keep the connection alive, this is probably
-the option you will want to use. The ``run_forever()`` function will automatically
-send a "pong" when it receives a "ping", per the specification.
+
+In this example, if a ping is received, and pong is sent as response, and then the
+client is notified, via ``on_ping()``.
+
+Further, a ping is transmitted every 60 seconds. If a pong is received, the client
+is notified, via ``on_pong()``. If no pong is received, within 10 seconds,
+``run_forever()`` will exit with an ``WebSocketTimeoutException``.
```
engn33r, line 460: nitpick: should be `a WebSocketTimeoutException`, not `an WebSocketTimeoutException`

Julian-O:
> Do you want to find a place to add a comment about the `recv_data_frame()` function being the source of the automatic pong response? That way we don't lose that valuable insight from this effort!

Good idea. Added comment to `recv_data_frame()` which is automatically reflected in documentation.

docs/source/examples.rst
Julian-O: Well spotted. Fixed.

PR #737, opened 2021-08-30
Add docstring to WebSocket.send_binary

### LINKED ISSUE #736: Document `WebSocket.send_binary` method?
The `WebSocket` class has a handy `send_binary()` method, however, it doesn't have a docstring and doesn't appear in the documentation, so I wasn't sure whether it was supposed to be part of the public API or not. (Though I notice there's at least one test for it.)

We should add a docstring either way, either to say "yes, here's what it does, use it" or "here's what it does, but it's deprecated, use `send(x, opcode=websocket.ABNF.OPCODE_BINARY)` instead".

I'm happy to submit a PR to add the docstring once I know which direction you want to go.

benhoyt:
Fixes #736

### DIFF
--- a/websocket/_core.py
+++ b/websocket/_core.py
@@ -314,6 +314,14 @@ class WebSocket(object):
         return length
 
     def send_binary(self, payload):
+        """
+        Send a binary message (OPCODE_BINARY).
+
+        Parameters
+        ----------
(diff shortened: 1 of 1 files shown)


engn33r:
Thanks!

PR #751, opened 2021-10-19
Use bytes literal

imba-tjd:


engn33r:
Thanks, I made a very small edit to mention UTF-8 in the comments but I agree with you that this is a good improvement for the code

PR #762, opened 2021-11-24
add support to pass custom ssl.SSLContext into sslopt param

### LINKED ISSUE #761: Allow custom SSL context
Currently the sslopt parameter of run_forever() and create_connection() allows passing in various common SSL options as a dict and then _wrap_sni_socket() builds out an ssl.SSLContext() based on those options. While this probably works fine for most simple requirements, it may have been a better design decision to just allow the SSL context to be passed in directly. There are a ton of knobs and dials on the ssl.SSLContext class and some very complex things can be done with it (and new features are added all the time).

For example, I have a requirement that the websocket client perform SSL validation and work with any valid public cert based on the system trusted root certificate store (the current default, which works fine there), but *also* trust a list of self-signed *private* certificate authorities for situations where the system moves behind specific corporate firewalls.

It's easy to construct such a context:

```python
context = ssl.create_default_context() # loads the default system CAs
context.load_verify_locations(capath=corporate_firewalls_ca_dir) # loads some additional ones too
```

There is no way to do such a thing with websocket-client, and there is no way to pass this custom context into it.

Obviously you don't want to break backwards compatibility, so it's too late to scrap the whole dict of ssl options thing, but I don't think the solution here is to keep expanding the available options that are passed into that dict. Let's instead add one final option to that dict - `context` which gives anyone the power to make their own custom SSL context - with any possible configuration - and pass that in directly.

grintor:
closes #761

### DIFF
--- a/websocket/_http.py
+++ b/websocket/_http.py
@@ -211,33 +211,35 @@ def _open_socket(addrinfo_list, sockopt, timeout):
 
 
 def _wrap_sni_socket(sock, sslopt, hostname, check_hostname):
-    context = ssl.SSLContext(sslopt.get('ssl_version', ssl.PROTOCOL_TLS_CLIENT))
+    context = sslopt.get('context', None)
+    if not context:
+        context = ssl.SSLContext(sslopt.get('ssl_version', ssl.PROTOCOL_TLS_CLIENT))
 
-    if sslopt.get('cert_reqs', ssl.CERT_NONE) != ssl.CERT_NONE:
-        cafile = sslopt.get('ca_certs', None)
-        capath = sslopt.get('ca_cert_path', None)
-        if cafile or capath:
-            context.load_verify_locations(cafile=cafile, capath=capath)
-        elif hasattr(context, 'load_default_certs'):
-            context.load_default_certs(ssl.Purpose.SERVER_AUTH)
-    if sslopt.get('certfile', None):
-        context.load_cert_chain(
-            sslopt['certfile'],
-            sslopt.get('keyfile', None),
-            sslopt.get('password', None),
-        )
-    # see
-    # https://github.com/liris/websocket-client/commit/b96a2e8fa765753e82eea531adb19716b52ca3ca#commitcomment-10803153
-    context.verify_mode = sslopt['cert_reqs']
-    if HAVE_CONTEXT_CHECK_HOSTNAME:
-        context.check_hostname = check_hostname
-    if 'ciphers' in sslopt:
-        context.set_ciphers(sslopt['ciphers'])
-    if 'cert_chain' in sslopt:
-        certfile, keyfile, password = sslopt['cert_chain']
-        context.load_cert_chain(certfile, keyfile, password)
-    if 'ecdh_curve' in sslopt:
-        context.set_ecdh_curve(sslopt['ecdh_curve'])
+        if sslopt.get('cert_reqs', ssl.CERT_NONE) != ssl.CERT_NONE:
+            cafile = sslopt.get('ca_certs', None)
+            capath = sslopt.get('ca_cert_path', None)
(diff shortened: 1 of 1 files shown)


grintor:
Not sure why the tests are failing, but I don't think it's my fault. This is a pretty minimal edit, only two lines and some whitespace.

engn33r:
Thanks for the PR. It looks like the build fails on the first `self.assertRaises()` test of the `testConnect` test case of test_http.py because `OSError` is triggered instead of the `ProxyTimeoutError` that was expected. It might take some time to figure out what's going on, and maybe the test case needs modification, but I'd like to resolve this before merging.

In case you aren't able to recreate (I haven't yet successfully recreated the issue locally), here's the build error trace:
```
=================================== FAILURES ===================================
_____________________________ HttpTest.testConnect _____________________________

    def connect(
        self,
        dest_host: str,
        dest_port: int,
        timeout: float = None,
        _socket=None,
    ) -> socket.socket:
    
        if timeout is None:
            timeout = DEFAULT_TIMEOUT
    
        self._dest_host = dest_host
        self._dest_port = dest_port
        self._timeout = timeout
    
        try:
            if _socket is None:
>               _socket = connect_tcp(
                    host=self._proxy_host,
                    port=self._proxy_port,
                    timeout=timeout,
                )

/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/python_socks/sync/_proxy.py:46: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

host = 'example.com', port = 8080, timeout = 1, local_addr = None

    def connect_tcp(
        host: str,
        port: int,
        timeout: float = None,
        local_addr: Optional[Tuple[str, int]] = None,
    ) -> socket.socket:
        address = (host, port)
    sock = ws_proxy.connect(hostname, port, timeout=proxy.proxy_timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def connect(
        self,
        dest_host: str,
        dest_port: int,
        timeout: float = None,
        _socket=None,
    ) -> socket.socket:
    
        if timeout is None:
            timeout = DEFAULT_TIMEOUT
    
        self._dest_host = dest_host
        self._dest_port = dest_port
        self._timeout = timeout
    
        try:
            if _socket is None:
                _socket = connect_tcp(
                    host=self._proxy_host,
                    port=self._proxy_port,
                    timeout=timeout,
                )
    
            self._stream = SyncSocketStream(_socket)
            self._negotiate()
            return _socket
        except socket.timeout as e:
            self._close()
            raise ProxyTimeoutError('Proxy connection timed out: {}'.format(self._timeout)) from e
        except OSError as e:
            self._close()
            msg = 'Could not connect to proxy {}:{} [{}]'.format(
                self._proxy_host,
                self._proxy_port,
                e.strerror,
            )
>           raise ProxyConnectionError(e.errno, msg) from e
E           python_socks._errors.ProxyConnectionError: [Errno 101] Could not connect to proxy example.com:8080 [Network is unreachable]

/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/python_socks/sync/_proxy.py:65: ProxyConnectionError
```

engn33r:
Somehow I already introduced the edit from @grintor a few minutes ago in 9d1285e and this PR ended up only including my fix for the CI issue. Not quite sure what I did there, but the suggested code is now in the _http.py file

grintor:
Cool, thanks @engn33r. Let me know when you make a new release in pypi and I'll do a pr to update the docs.

Issue #756, opened 2021-11-13
Switch from `scripts` to `entry_points: console_scripts `


BastianZim:
Hi,

I was wondering if it might make sense to switch from `scripts` to `entry_points console_scripts ` in _setup.py_. This is the more modern and preferred approach and recommended by PyPA nowadays.

Example:
```
entry_points={
    'console_scripts': [
        'sample=sample:main',
    ],
},
```

Reference: https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts
Code: https://github.com/websocket-client/websocket-client/blob/cf0eb68a63747222434e382e24ea38836201dc30/setup.py#L71

engn33r:
I spent some time trying to update this but I'm getting `ModuleNotFoundError` errors with the newer `entry_points` approach, even when moving the bin directory into the websocket directory. I would be happy to accept a PR to switch this.

BastianZim:
Hi @engn33r thanks for starting this. I currently don't have but hope to get some time to work on this during the next few weeks. Just to check beforehand because I couldn't find anything: Is this part also covered by the tests or should some for that part also be added?

engn33r:
The [build.yml GitHub action](https://github.com/websocket-client/websocket-client/blob/master/.github/workflows/build.yml) tests the build process (currently only on ubuntu), which includes the creation of wsdump.py, but the code coverage tests only covers the `websocket` directory and therefore does not test the wsdump.py script in the `bin` directory.

BastianZim:
I had some time sooner than I thought, PR is in https://github.com/websocket-client/websocket-client/pull/763

PR #763, opened 2021-11-28
Add entry points

BastianZim:
Closes https://github.com/websocket-client/websocket-client/issues/756

This PR adds `entry_points` as defined in the issue. Additionally, the CLI tool is tested by running the help command as it is currently not tested.

Note: It might make sense to replace this with proper tests at some point.


engn33r:
Thanks, this is definitely better than what I attempted. Currently CI is giving a `wsdump: command not found` error with the Ubuntu build process. I'll test this out locally later to see if I can recreate. I will also need to check why code coverage dropped 8.61%, but it's probably because wsdump.py was moved into the websocket/ directory.

BastianZim:
> Currently CI is giving a wsdump: command not found error with the Ubuntu build process.

I think that's my bad. I didn't check if the package is actually installed which seems to not be the case. To be honest, running `-h` isn't a good test anyway so I think it'd be better to remove it again. The only reason I didn't put in a full test is that I'm not familiar enough with pytest to know how to test CLI commands. If you know more about that, please feel free to push to my branch directly.
The command works locally for me so it should be just about the tests, not the PR.

> I will also need to check why code coverage dropped 8.61%, but it's probably because wsdump.py was moved into the websocket/ directory.

Yeah, I think it's just the change of location. Previously, only websocket was included.

BastianZim:
I have moved everything under the normal package and just made the setup slightly leaner, which should also make the tests easier.

Based on other packages I guess the easiest would be to call `subprocess.run()` in a test file with the CLI and test a couple of commands. What do you think?
I don't have any experience with unit tests as I usually use pytest but please feel free to push to my branch directly, if you have some ideas.

engn33r:
I sent a PR to suggest two changes to fix the handling of GitHub Actions CI:
1. Test the wsdump.py script with `wsdump.py -h` (I tried `wsdump -h` but CI required the .py extension) right after websocket-client is installed. websocket-client currently doesn't get installed in the build.yml test, so I put this check in codecoverage.yml
2. Omit the wsdump.py script from code coverage because wsdump.py doesn't contribute any new functionality to the underlying websocket-client library. As long as the `wsdump.py -h` command doesn't trigger an error, the script should be installed. I'm not too concerned about code coverage for wsdump.py because most users of websocket-client are importing the library for custom purposes.

BastianZim:
Ok makes all sense. Let's see if it works this time.

engn33r:
Although the `wsdump.py -h` test worked when I ran it in the GitHub CI of a test fork yesterday, it is not working today. I am going to remove the wsdump.py test entirely from CI since it's not critical to the overall functionality of websocket-client as a library.

engn33r:
CI is passing now. I will merge this PR after 1-2 days in case you have any comments or final edits

BastianZim:
Ok, thanks. I'll just test this locally again as it's currently not tested at all - just to verify that everything's working as intended and then would post the result here.

BastianZim:
Tested it locally and `wsdump -h` works, would be good to go from my side.

engn33r:
Great, thanks for the PR!

BastianZim:
Thanks for merging!

PR #770, opened 2021-12-07
rel example

bubbleboy14:
simple example of using rel dispatcher in applications that require multiple WebSocketApp instances

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -663,6 +663,24 @@ by `issue #351 <https://github.com/websocket-client/websocket-client/issues/351>
 
 `Work in progress - coming soon`
 
+Dispatching Multiple WebSocketApps
+==================================
+
+You can use an asynchronous dispatcher such as `rel <https://pypi.org/project/rel/>`_ to run multiple WebSocketApps in the same application without resorting to threads.
+
+::
+
+  import websocket, rel
+
+  addr = "wss://api.gemini.com/v1/marketdata/%s"
+
+  rel.safe_read()
+  for symbol in ["BTCUSD", "ETHUSD", "ETHBTC"]:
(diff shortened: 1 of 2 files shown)


engn33r:
Thanks a bunch for the PR, since I haven't used a dispatcher parameter for `run_forever()` before. However, could we put this example in the docs/source/examples.rst file? I am happy to do so if you don't mind my closing this PR and submitting a separate commit with your code example.

bubbleboy14:
I just added an examples.rst entry - is that sufficient?

engn33r:
That looks great, thank you! I might remove the examples/rel_client.py file before I merge in a few days if that's ok with you, to avoid duplication.

bubbleboy14:
sounds good :)

engn33r:
@bubbleboy14 when I try to install rel with `pip3 install rel`, my system does not find rel 0.4.6 that I see on PyPi and instead it tries to install rel 0.3.1 (unsuccessfully). In fact, the PyPi egg for Python 3.9 has an error in the Python version column, which probably explains why my system's Python 3.9 cannot find it: https://pypi.org/project/rel/#files

When I use `twine check rel-0.4.6-py3.9.egg` on the Python 3.9 file, I get an error, so I think the Python 3.9 (and possibly other) PyPi files need to be rebuilt since I think this error exists for Python 3.9 for all releases since 0.3.1. I guess I should have tested the example myself before merging :)

bubbleboy14:
@engn33r thank you so much for pointing this out! Good call - it looks like the 3.9 eggs aren't versioned right for some reason. Hah, I guess everyone's using wheels now instead of eggs?

I just uploaded a wheel and tested installation on some server, works great. LMK if it's still busted on your end, and thanks again for the heads up!

engn33r:
Much better, the example works for me now! I even tested that the websocket keeps working normally after the network is temporarily disconnected. Thank you for fixing

PR #771, opened 2021-12-17
Add missing CLOSE opcodes

Pithikos:
As registered here: https://www.iana.org/assignments/websocket/websocket.xhtml


engn33r:
Thank you for the PR

Issue #779, opened 2022-01-19
echo.websocket.org service no longer available


SidneyAllen:
I noticed that Kaazing's websocket.org echo URL is used in documentation and examples for this repo.
https://github.com/websocket-client/websocket-client/search?q=echo.websocket.org

The company that acquired Kaazing shut down the site sometime last year.  For a hackathon project at my company I stood up a replacement on heroku at http://echo.websocket.events/.ws. Got approval to keep it running for other developers to use.

I'm happy to make a PR and replace echo.websockets.org.  Also I wrote up a blog post explaining to others how to deploy their own echo server on Heroku if you wanna try that. https://www.lob.com/blog/websocket-org-is-down-here-is-an-alternative

Cheers!

engn33r:
Yes! A PR with the new link would be excellent. I had many issues with echo.websockets.org not "echoing" properly throughout 2021 so I'm not surprised to see it's shutting down.

engn33r:
@SidneyAllen FYI I noticed websocket.events is not resolving. It might be related to domain renewal, I didn't look too closely.

senechko:
hi @engn33r this has been addressed and it is now working again. thank you for reporting this.

PR #781, opened 2022-01-21
Update the echo.websocket URL in README and 2 files in examples folder

SidneyAllen:
echo.websocket.org is no longer available, so replacing with an alternate server that will echo back whatever you send via websockets.

I explored the use of echo.websocket.org in examples.rst and they go beyond a basic echo demonstrate proxying, custom headers, and other advanced use cases.

Since, echo.websocket.events is currently just a basic echo-server it doesn't seem appropriate for those use cases.  For this reason, I didn't modify any of the examples in the docs folder.

Hopefully the community finds this change useful 

Reference issue #779

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,8 @@ server is running and responds properly to a specific request.
 ```python
 from websocket import create_connection
 
-ws = create_connection("ws://echo.websocket.org/")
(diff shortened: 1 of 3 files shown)


engn33r:
Thank you for the PR!

PR #788, opened 2022-02-07
fix!: `WebSocketApp.run_forever()` returning `None` after `close()` is called

### LINKED ISSUE #785: Clarify documentation for teardown return from run_forever()
The [doc-comment](https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L258-L261) in `run_forever` says:

```
teardown: bool
            False if caught KeyboardInterrupt, True if other exception was raised during a loop
```

This doesn't cover are the cases where `run_forever` returns `None`.

One case is on signal `SIGTERM` which causes `keep_running` to become `False` and therefore [SSLDispatcher.read](https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L66) returns `None`.  Pressing the stop button in PyCharm will trigger this.

I _think_ the intention here is to allow the caller to differentiate between a graceful and non-graceful exit.  Given that it may be sensible to have `run_forever` always return either `False` or `True` and have both `SIGINT` (i.e. `KeyboardInterrupt` exception) and other non-exceptional status (i.e. `SIGTERM`) return `False`.

fujiapple852:
Closes #785 

BREAKING CHANGE: the return value of `WebSocketApp.run_forever()` will no longer return the (undocumented) `None` value.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -258,7 +258,8 @@ class WebSocketApp:
         Returns
         -------
         teardown: bool
-            False if caught KeyboardInterrupt, True if other exception was raised during a loop
+            False if the `WebSocketApp` is closed or caught KeyboardInterrupt,
+            True if any other exception was raised during a loop.
         """
 
         if ping_timeout is not None and ping_timeout <= 0:
@@ -367,6 +368,7 @@ class WebSocketApp:
                 return True
 
             dispatcher.read(self.sock.sock, read, check)
+            return False
         except (Exception, KeyboardInterrupt, SystemExit) as e:
             self._callback(self.on_error, e)
             if isinstance(e, SystemExit):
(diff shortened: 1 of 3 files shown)


engn33r:
The CI output shows this breaks two tests in websocket/tests/test_websocket.py: the `testIter` test and the `testNext` test. It could be because the tests are poorly written, but I don't have time to investigate the exact cause right now.

fujiapple852:
@engn33r it looks like `wss://stream.meetup.com/2/rsvps` no longer exists (returns 404).  I replaced this with `wss://api.bitfinex.com/ws/2` in these two tests (and one other, which wasn't strictly necessary) which allows them to pass locally.

fujiapple852:
In passing, I spotted that `testRunForeverDispatcher` passes but triggers a (silently ignored) error because of a seemingly [bogus](https://github.com/websocket-client/websocket-client/blob/master/websocket/tests/test_app.py#L90) call to `WebSocketApp.recv()`.  I haven't pushed a change for this as there are a few other oddities with this test and `testKeepRunning` that I don't understand so is probably worth a separate issue.

fujiapple852:
I added two new test cases to cover the graceful and exceptional shutdown of `run_forever()`.  They are both triggered by a timer so could be racy (`0.2` secs for now)

engn33r:
Thanks for the PR, and for fixing the broken WebSocket endpoint URLs!

Issue #786, opened 2022-02-04
gets disconnected when recieving HIGH number of msgs per second


pssolanki111:
Hey There,

Appreciate all the work creating this amazing client implementation. 

I used this to write the callback based streaming client in my [polygon API package](https://github.com/pssolanki111/polygon) and it has been doing very good work.

A few days back, one of the users of my lib reported disconection issues when the number of msgs being recieved per second was high. I made sure they weren't doing any heavy lifting in the handler itself. In fact they were just pushing messages to a redis queue from threads 

Now on to some pointers, 

-  I'm aware of [the FAQ on why the lib is slow](https://websocket-client.readthedocs.io/en/latest/faq.html#why-is-this-library-slow). about that though, we are not sending more than a couple of msgs. All the volume is being received. Further, I keep utf-8 validation disabled.
- another accompanying issue is the fact that implementing a reliable **reconection mechanism** for this client hasn't been the easiest thing to do. Every approach has some quirks (unclosed clients, recursive calls from close handler, messed up threading to name a few)

What could be the reasons behind why it's not able to process the messages quickly enough and what could be done to resolve it?

further have  we found a reliable and robust way to reconnect in case it does disconnect? @engn33r 

Thanks : )

engn33r:
I don't see any error messages related to your issue which limits how much help I can provide. To handle large numbers of connections on Linux, which is probably what is used if redis is involved, you should make sure the value of `ulimit -n` is large, which is described in #607. That same issue mentions creating a custom dispatcher, which is what I would suggest for handling the reconnection process. An easier way to handle the reconnections is using rel, which is in [the README](https://github.com/websocket-client/websocket-client#long-lived-connection) and was the best solution I found for #580. The [threading doc](https://websocket-client.readthedocs.io/en/latest/threading.html) mentions other ideas.

pssolanki111:
> To handle large numbers of connections on Linux, which is probably what is used if redis is involved,

No it happens on windows as well (yes redis can run on windows) as *nix OS. Windows doesn't impose a file dscriptor limit on processes. 

> I don't see any error messages related to your issue which limits how much help I can provide

Oh yeah My bad. I had the error messages copied, forgot to append to msg lol. here it is (i only happen to have a picture at the moment, it's a ping/pong time out)
[...]

PR #795, opened 2022-02-13
custom dispatcher pings/timeouts

bubbleboy14:
Two things:

1) _socket.recv() error handling
  a) fixed socket.error except logic (_socket.py line 95) - I believe the line substituted fulfills the intention of the three lines replaced
  b) throw WebSocketTimeoutException on TimeoutError

2) custom dispatchers (_app.WrappedDispatcher)
  a) fixes pyevent integration - calls read() w/ correct arguments
  b) simplifies rel integration - no need for rel.safe_read()
  c) supports ping_timeout/ping_interval via dispatcher.timeout() w/ check_callback(), if ping_timeout specified

So these changes should make the two modes (standard and w/ custom dispatcher) functionally equivalent in these ways. That is, if you decide to start or stop using a custom dispatcher, your ping_timeout/ping_interval handling won't change (check() function in _app is still called), and your application won't have to handle any new exceptions (such as TimeoutError, which is now expressed as WebSocketTimeoutException).

### DIFF
--- a/.github/workflows/codecoverage.yml
+++ b/.github/workflows/codecoverage.yml
@@ -33,12 +33,22 @@ jobs:
           LOCAL_WS_SERVER_PORT: 8765
       - name: Run SSL test with extra environment variable and create report
         run: |
+          ls $WEBSOCKET_CLIENT_CA_BUNDLE
+          python3 -c "import ssl; print(ssl.get_default_verify_paths().capath)"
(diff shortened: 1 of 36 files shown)


pssolanki111 (approved):
the changes sound good as discussed in #786

engn33r:
Thanks for the PR!

PR #804, opened 2022-03-19
Add support to pass pre-initialized stream socket when creating WebSocketApp (#802)

### LINKED ISSUE #802: Passing pre-initialized stream socket when creating WebSocketApp
`WebSocket.connect()` supports [passing a pre-initialized stream socket](https://github.com/websocket-client/websocket-client/blob/master/websocket/_core.py#L240). But looks like `WebSocketApp` doesn't accept `socket` argument when initializing. Currently I have to modify this library locally, but I'd love to have this feature officially.
  
Could you please add it? Thanks a lot.

redphx:
Closes #802

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -185,7 +185,6 @@ class WebSocketApp:
         self.last_ping_tm = 0
         self.last_pong_tm = 0
         self.subprotocols = subprotocols
-        
         self.prepared_socket = socket
 
     def send(self, data, opcode=ABNF.OPCODE_TEXT):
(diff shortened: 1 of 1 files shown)


websocket/_app.py
```diff
@@ -182,6 +185,8 @@ def __init__(self, url, header=None,
         self.last_ping_tm = 0
         self.last_pong_tm = 0
         self.subprotocols = subprotocols
+        
+        self.prepared_socket = socket
```
redphx, line 189: I didn't use `self.socket` because we're already had `self.sock`, so I went with `self.prepared_socket`.

engn33r:
Thanks, this looks like a nice improvement!

redphx:
Hi @engn33r, thanks for accepting my MR! Any change for an official release?  
I'm using this in my `requirements.txt` file:  
```
git+https://github.com/websocket-client/websocket-client@1185fea#egg=websocket-client
```
But doing so requires normal users to have `git` installed, and I think that's unnecessary to most of them.  
Thank you.

PR #826, opened 2022-05-26
Fix unclosed socket when 'error.errno in eConnRefused'.

bvd0:


engn33r:
Thanks for the PR. Can you provide more information about why this fix is necessary so I can understand the fix before merging? What conditions caused the error message you encountered? Is there a way to replicate the issue? Can you share a [debug trace](https://websocket-client.readthedocs.io/en/latest/examples.html#debug-and-logging-options) of this issue?

bvd0:
This is actually more of an enhancement than a fix.

If you run python with `-Walways` and `-X tracemalloc`, there is a `ResourceWarning` every time the connection is refused.
This example command should cause the `ResourceWarning`:
```sh
python3 -Walways -X tracemalloc -c "import websocket ; websocket.create_connection('ws://localhost:65535')"  # ('65535' can be any port that has nothing listening on it)
```

Using `except AttributeError:` instead of `except:` prevents other exceptions such as `KeyboardInterrupt` from being accidentally caught.

engn33r:
Nice catch, I can recreate the difference before/after this PR

PR #838, opened 2022-07-11
reconnect

bubbleboy14:
run_forever() has a new kwarg, "reconnect".

If "reconnect" is True (which is the default) and the connection ends (or can't be established), it waits for 5 seconds and tries again.

This has been tested with run_forever() and run_forever(dispatcher=rel), and seems to work pretty great.

LMK what you think!

NB: I think I merged in upstream weirdly (maybe forgot to pull first), so there are lots of commits listed going back 7 months. However, the only actual changes (with one small exception - see diff) are in run_forever().

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -30,15 +30,19 @@ limitations under the License.
 
 __all__ = ["WebSocketApp"]
 
-
-class Dispatcher:
+class DispatcherBase:
     """
-    Dispatcher
+    DispatcherBase
     """
     def __init__(self, app, ping_timeout):
         self.app = app
         self.ping_timeout = ping_timeout
 
+
+class Dispatcher(DispatcherBase):
+    """
+    Dispatcher
+    """
     def read(self, sock, read_callback, check_callback):
         while self.app.keep_running:
             sel = selectors.DefaultSelector()
@@ -51,15 +55,15 @@ class Dispatcher:
             check_callback()
             sel.close()
 
+    def timeout(self, seconds, callback):
+        time.sleep(seconds)
+        callback()
 
-class SSLDispatcher:
+
+class SSLDispatcher(DispatcherBase):
     """
     SSLDispatcher
     """
-    def __init__(self, app, ping_timeout):
-        self.app = app
-        self.ping_timeout = ping_timeout
-
     def read(self, sock, read_callback, check_callback):
         while self.app.keep_running:
             r = self.select()
@@ -95,6 +99,9 @@ class WrappedDispatcher:
     def read(self, sock, read_callback, check_callback):
         self.dispatcher.read(sock, read_callback)
         self.ping_timeout and self.dispatcher.timeout(self.ping_timeout, check_callback)
+
+    def timeout(self, seconds, callback):
+        self.dispatcher.timeout(seconds, callback)
 
 
 class WebSocketApp:
@@ -390,15 +397,10 @@ class WebSocketApp:
                 raise
             if reconnect and not isinstance(e, KeyboardInterrupt):
                 _logging.warning("websocket disconnected (retrying in %s seconds) [%s frames in stack]"%(reconnect, len(inspect.stack())))
-                if custom_dispatcher:
-                    custom_dispatcher.timeout(reconnect, setSock)
-                else:
-                    time.sleep(reconnect)
-                    setSock()
+                dispatcher.timeout(reconnect, setSock)
             else:
                 teardown()
 
-        custom_dispatcher = dispatcher
         dispatcher = self.create_dispatcher(ping_timeout, dispatcher, not not sslopt)
 
         if ping_interval:
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -317,7 +317,7 @@ class WebSocketApp:
             # Finally call the callback AFTER all teardown is complete
             self._callback(self.on_close, close_status_code, close_reason)
 
-        def setSock():
+        def setSock(doread=False):
             self.sock = WebSocket(
                 self.get_mask_key, sockopt=sockopt, sslopt=sslopt,
                 fire_cont_frame=self.on_cont_message is not None,
@@ -335,6 +335,8 @@ class WebSocketApp:
 
                 self._callback(self.on_open)
 
+                _logging.warning("websocket connected")
+                doread and dispatcher.read(self.sock.sock, read, check)
             except ConnectionRefusedError as e:
                 handleDisconnect()
 
@@ -384,9 +386,7 @@ class WebSocketApp:
             if reconnect:
                 _logging.warning("websocket disconnected (retrying in 5 seconds)")
                 time.sleep(5)
-                setSock()
-                _logging.warning("websocket reconnected")
-                dispatcher.read(self.sock.sock, read, check)
+                setSock(True)
                 return True
 
         dispatcher = self.create_dispatcher(ping_timeout, dispatcher, not not sslopt)
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -37,21 +37,39 @@ __all__ = ["WebSocketApp"]
 class Dispatcher:
     """
     Dispatcher
-    Note: read() removed along with keep_running variable
     """
     def __init__(self, app, ping_timeout):
         self.app = app
         self.ping_timeout = ping_timeout
+
+    def read(self, sock, read_callback, check_callback):
+        while self.app.keep_running:
+            sel = selectors.DefaultSelector()
+            sel.register(self.app.sock.sock, selectors.EVENT_READ)
+
+            r = sel.select(self.ping_timeout)
+            if r:
+                if not read_callback():
+                    break
+            check_callback()
+            sel.close()
 
 
 class SSLDispatcher:
     """
     SSLDispatcher
-    Note: read() removed along with keep_running variable
     """
     def __init__(self, app, ping_timeout):
         self.app = app
         self.ping_timeout = ping_timeout
+
+    def read(self, sock, read_callback, check_callback):
+        while self.app.keep_running:
+            r = self.select()
+            if r:
+                if not read_callback():
+                    break
+            check_callback()
 
     def select(self):
         sock = self.app.sock.sock
@@ -77,7 +95,7 @@ class WebSocketApp:
                  on_open=None, on_message=None, on_error=None,
                  on_close=None, on_ping=None, on_pong=None,
                  on_cont_message=None,
-                 get_mask_key=None, cookie=None,
+                 keep_running=True, get_mask_key=None, cookie=None,
                  subprotocols=None,
                  on_data=None):
         """
@@ -126,6 +144,8 @@ class WebSocketApp:
             The 2nd argument is utf-8 string which we get from the server.
             The 3rd argument is data type. ABNF.OPCODE_TEXT or ABNF.OPCODE_BINARY will be came.
             The 4th argument is continue flag. If 0, the data continue
+        keep_running: bool
+            This parameter is obsolete and ignored.
         get_mask_key: function
             A callable function to get new mask keys, see the
             WebSocket.set_mask_key's docstring for more information.
@@ -146,6 +166,7 @@ class WebSocketApp:
         self.on_ping = on_ping
         self.on_pong = on_pong
         self.on_cont_message = on_cont_message
+        self.keep_running = False
         self.get_mask_key = get_mask_key
         self.sock = None
         self.last_ping_tm = 0
@@ -173,6 +194,7 @@ class WebSocketApp:
         """
         Close websocket connection.
         """
+        self.keep_running = False
         if self.sock:
             self.sock.close(**kwargs)
             self.sock = None
@@ -252,6 +274,7 @@ class WebSocketApp:
         if self.sock:
             raise WebSocketException("socket is already opened")
         thread = None
+        self.keep_running = True
         self.last_ping_tm = 0
         self.last_pong_tm = 0
 
@@ -269,6 +292,7 @@ class WebSocketApp:
             if thread and thread.is_alive():
                 event.set()
                 thread.join()
+            self.keep_running = False
             if self.sock:
                 self.sock.close()
             close_status_code, close_reason = self._get_close_args(
@@ -305,6 +329,9 @@ class WebSocketApp:
                 thread.start()
 
             def read():
+                if not self.keep_running:
+                    return teardown()
+
                 op_code, frame = self.sock.recv_data_frame(True)
                 if op_code == ABNF.OPCODE_CLOSE:
                     return teardown(frame)
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -384,7 +384,7 @@ class WebSocketApp:
 
         def handleDisconnect():
             if reconnect:
-                _logging.warning("websocket disconnected (retrying in 5 seconds)")
+                _logging.warning("websocket disconnected (retrying in %s seconds)"%(reconnect,))
                 time.sleep(reconnect)
                 setSock(True)
                 return True
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -1,3 +1,4 @@
+import inspect
 import selectors
 import sys
 import threading
@@ -317,7 +318,7 @@ class WebSocketApp:
             # Finally call the callback AFTER all teardown is complete
             self._callback(self.on_close, close_status_code, close_reason)
 
-        def setSock(doread=False):
+        def setSock():
             self.sock = WebSocket(
                 self.get_mask_key, sockopt=sockopt, sslopt=sslopt,
                 fire_cont_frame=self.on_cont_message is not None,
@@ -336,9 +337,9 @@ class WebSocketApp:
                 self._callback(self.on_open)
 
                 _logging.warning("websocket connected")
-                doread and dispatcher.read(self.sock.sock, read, check)
-            except ConnectionRefusedError as e:
-                handleDisconnect()
+                dispatcher.read(self.sock.sock, read, check)
+            except (Exception, ConnectionRefusedError, KeyboardInterrupt, SystemExit) as e:
+                handleDisconnect(e)
 
         def read():
             if not self.keep_running:
@@ -348,7 +349,7 @@ class WebSocketApp:
                 op_code, frame = self.sock.recv_data_frame(True)
             except WebSocketConnectionClosedException as e:
                 _logging.error("WebSocketConnectionClosedException - %s"%(reconnect and "reconnecting" or "goodbye"))
-                return handleDisconnect()
+                return handleDisconnect(e)
             if op_code == ABNF.OPCODE_CLOSE:
                 return teardown(frame)
             elif op_code == ABNF.OPCODE_PING:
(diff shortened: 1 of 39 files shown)


websocket/_app.py
```diff
@@ -385,7 +385,7 @@ def check():
         def handleDisconnect():
             if reconnect:
                 _logging.warning("websocket disconnected (retrying in 5 seconds)")
```
brunchboy, line 387: I’d think you’d want the value of `reconnect` in this log message, rather than leaving it hard-coded at 5?

bubbleboy14: yes indeed! very good call :)

engn33r:
@maksimu I'm away from my computer right now, but later this month I plan to catch up on everything

engn33r:
Great upgrade, thanks @bubbleboy14! My comments:
- Nice idea to use inheritance for the Dispatcher classes to reduce redundancy
- I like how the logging message mentions how many frames are on the stack, which gives an idea of how long the disconnection has lasted
- Good idea replacing the teardown try/except with functions to improve function call control

I found no issues when testing these changes
1. on an encrypted wss endpoint on the web where I stopped/restarted my network connection several times
2. on a local unencrypted ws server where I stopped/restarted the server process several times

To err on the cautious side, I will make a new 3.4.0 release for this in case the upgrades have any unexpected side effects that I didn't catch.

For future reference, the CI linting error exists in the master branch and was not introduced by this PR.

NadavK:
Firstly, thanks for this important upgrade!
Please note that for me, when the server-side raises an exception, the **client** terminates.

For instance, changing the ```echo-server.py``` app to raise an exception, e.g. ```await websocket.send(XXX)``` expectedly fails ```with name 'XXX' is not defined```
But interestingly, also the client fails:

```
Running forever...
--- request header ---
GET / HTTP/1.1
Upgrade: websocket
Host: localhost:8765
Origin: http://localhost:8765
Sec-WebSocket-Key: 7auS5TiZnKAdDwJTNP883Q==
Sec-WebSocket-Version: 13
Connection: Upgrade


-----------------------
--- response header ---
HTTP/1.1 101 Switching Protocols
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Accept: CxFIBMn0pIottRDSWg+4kkuBk1g=
Date: Thu, 25 Aug 2022 14:27:55 GMT
Server: Python/3.10 websockets/10.3
-----------------------
websocket connected
++Sent raw: b'\x81\x87ft\xcbI\x0b\x11\xb8:\x07\x13\xae'
++Sent decoded: fin=1 opcode=1 data=b'message'
++Rcv raw: b'\x88\x02\x03\xf3'
++Rcv decoded: fin=1 opcode=8 data=b'\x03\xf3'
++Sent raw: b'\x88\x82 \xcbP;##'
++Sent decoded: fin=1 opcode=8 data=b'\x03\xe8'
### closed ###: 1011
Thread terminating...
```

here is my client code:
```python
def on_message(self, message):
    print(message)

def on_error(self, error):
    print("### error ###:", error)

def on_close(self, close_status_code, close_msg):
    print("### closed ###:", close_status_code, close_msg)

def send(self, message):
    print("### sending message ###")
    self.send(self, message)

if __name__ == '__main__':
    websocket.enableTrace(True)
    ws = websocket.WebSocketApp(
        'ws://localhost:8765/',
        on_message=on_message,
        on_error=on_error,
        on_close=on_close)

    scheduler = BackgroundScheduler(daemon=False)
    scheduler.add_job(ws.send, 'interval', seconds=2, args=['message'])
    scheduler.start()

    print("Running forever...")
    ws.run_forever()
    print("Thread terminating...")
```

bubbleboy14:
Hello @NadavK. Hm, you're talking about this one, right:

https://github.com/websocket-client/websocket-client/blob/master/websocket/tests/echo-server.py

On your client, are you using the BackgroundScheduler in PyBackground (https://pypi.org/project/PyBackground/) or apscheduler (https://apscheduler.readthedocs.io/en/stable/userguide.html) or some other project? Could you please show us your full client code, or at least the import lines?

Thanks!

engn33r:
I can recreate what @NadavK shared. My opinion is that this is expected behavior and I [updated the README](https://github.com/websocket-client/websocket-client/commit/cc09510e594c5d7ebe311098982429d6b19d723b) to clarify this detail. I will share my explanation and then the code I used to recreate.

The "client fails" action happens when the `on_close` callback in the client code is entered. This means the connection has been closed by the server and the client received close code 1011, which [is for STATUS_UNEXPECTED_CONDITION](https://github.com/websocket-client/websocket-client/blob/0c9ea4aa92aad5cf15a0407a7b106c9f7d4625c8/websocket/_abnf.py#L78). This happened because the server outputs the error `NameError: name 'XXX' is not defined`, and because the server is in an unexpected condition, it closes the connection. So if the server closes the connection, the client cannot automatically reconnect because the connection is closed, so should the client create a new connection to the server? If the server closes the connection, `run_forever` does not automatically attempt to create a new connection to the server by default, but adding the line `ws.run_forever(dispatcher=rel)` into the `on_close` callback will make the server reconnect when the connection is closed. In this specific example, it will cause an infinite loop, which is not ideal. I like the current way websocket-client handles this because a server that closes the connection and sends a close status code normally has a reason for doing so, meaning custom logic should be added in `on_close` depending on how the server normally responds and what the client wants to do. If the client or server drops offline without closing the connection, `run_forever` should attempt to reconnect because the connection is still open because no close status code was sent or received.

**Server code** (lightly modified from websocket/tests/echo-server.py in this repo)
```
#!/usr/bin/env python

# From https://github.com/aaugustin/websockets/blob/main/example/echo.py

import asyncio
import websockets
import os

LOCAL_WS_SERVER_PORT = os.environ.get('LOCAL_WS_SERVER_PORT', '8765')


async def echo(websocket, path):
    async for message in websocket:
        await websocket.send(XXX)


async def main():
    async with websockets.serve(echo, "localhost", LOCAL_WS_SERVER_PORT):
        await asyncio.Future()  # run forever

asyncio.run(main())
```

**Client code** (lightly modified from Long-Lived Connection code)
```
import websocket
import _thread
import time
import rel

def on_message(ws, message):
    print(message)

def on_error(ws, error):
    print(error)

def on_close(ws, close_status_code, close_msg):
    print("### closed ###")
    print("close_status_code: " + str(close_status_code))
    print("close_msg: " + close_msg)

def on_open(ws):
    print("Opened connection")
    ws.send("hello")

if __name__ == "__main__":
    websocket.enableTrace(True)
    ws = websocket.WebSocketApp("ws://127.0.0.1:8765",
                              on_open=on_open,
                              on_message=on_message,
                              on_error=on_error,
                              on_close=on_close)

    ws.run_forever(dispatcher=rel)  # Set dispatcher to automatic reconnection
    rel.signal(2, rel.abort)  # Keyboard Interrupt
    rel.dispatch()
```

mkapuza:
@engn33r I agree with you logic here and believe this is the right decision. 

Do you have a suggestion for an exponential reconnection that doesn't result in a potential stack overflow like calling `ws.run_forever(dispatcher=rel)` in the `on_close` method?

brunchboy:
In my case, I control the server and the client, and my absolute requirement is that no matter what, the client should retry its connection if it goes away for any reason. Because of that, I decided to put that logic outside of websocket-client, in the launchd configuration which runs it, to restart it whenever the process exits. The discussion above reinforces my plan to stick with the older version of websocket-client, and leave the reconnection outside of it.

bubbleboy14:
Hello all! @engn33r , @brunchboy , @mkapuza , check out this pull request:

https://github.com/websocket-client/websocket-client/pull/854

I adjusted the synchronous reconnect logic to avoid growing the stack. Thoughts?

aantn:
Hi, we're seeing strange behaviour since updating clients to 1.4.0.

On the websocket **server** (which uses a different package) the number of open websocket clients seems to infinitely increase. That could be due to a server bug, but I'm wondering if the change in the client behaviour could be playing a role in triggering it?

bubbleboy14:
Hey @aantn, https://github.com/websocket-client/websocket-client/pull/854 should fix this. Could you please test and let us know?

@engn33r, maybe we should change the "reconnect" kwarg default to 0 to preserve the previous behavior for existing integrations - what do you think?

milosivanovic:
There seems to be an issue with `rel` - it's extremely inefficient. Instead of blocking and waiting for data, it's polling every 0.3ms:

```
[root@machine ~]# strace -p 12345
strace: Process 12345 attached
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
poll([{fd=22, events=POLLIN}], 1, 0)    = 0 (Timeout)
pselect6(0, NULL, NULL, NULL, {tv_sec=0, tv_nsec=300000}, NULL) = 0 (Timeout)
```

You can check this by running `strace -p <pid_of_process>` against the PID of the Python process where the websocket client is implemented with the `rel` dispatcher.

Is there an alternative to `rel`? I don't think we should be suggesting this solution based on the inefficient busy-waiting it's doing.

bubbleboy14:
@milosivanovic the main alternative is to run without an async dispatcher, which is the default. If you want an alternative async dispatcher, pyevent has the same API, but I'm not sure if it runs on Python 3.

Bear in mind that the sole purpose of an asynchronous dispatcher is to run multiple WebSocketApps in the same process. If you don't have this requirement, the default (synchronous) dispatcher is all you'll ever need.

milosivanovic:
@bubbleboy14 I see. In the README for this project, it suggests that if we want automatic reconnection, we should use the `rel` dispatcher. Did I misunderstand this? Are you suggesting that the client will still try to automatically reconnect without any loss in functionality if we omit `rel`? If so, would it be helpful if I opened an issue to have that text clarified?

bubbleboy14:
@milosivanovic for that, use the "reconnect" kwarg, eg:

ws.run_forever(reconnect=5)

In the above example, 5 is the reconnect interval, so if the connection drops, it will be re-established 5 seconds later.

pouya817:
> I can recreate what @NadavK shared. My opinion is that this is expected behavior and I [updated the README](https://github.com/websocket-client/websocket-client/commit/cc09510e594c5d7ebe311098982429d6b19d723b) to clarify this detail. I will share my explanation and then the code I used to recreate.
> 
> The "client fails" action happens when the `on_close` callback in the client code is entered. This means the connection has been closed by the server and the client received close code 1011, which [is for STATUS_UNEXPECTED_CONDITION](https://github.com/websocket-client/websocket-client/blob/0c9ea4aa92aad5cf15a0407a7b106c9f7d4625c8/websocket/_abnf.py#L78). This happened because the server outputs the error `NameError: name 'XXX' is not defined`, and because the server is in an unexpected condition, it closes the connection. So if the server closes the connection, the client cannot automatically reconnect because the connection is closed, so should the client create a new connection to the server? If the server closes the connection, `run_forever` does not automatically attempt to create a new connection to the server by default, but adding the line `ws.run_forever(dispatcher=rel)` into the `on_close` callback will make the server reconnect when the connection is closed. In this specific example, it will cause an infinite loop, which is not ideal. I like the current way websocket-client handles this because a server that closes the connection and sends a close status code normally has a reason for doing so, meaning custom logic should be added in `on_close` depending on how the server normally responds and what the client wants to do. If the client or server drops offline without closing the connection, `run_forever` should attempt to reconnect because the connection is still open because no close status code was sent or received.
> 
> **Server code** (lightly modified from websocket/tests/echo-server.py in this repo)
> 
> ```
> #!/usr/bin/env python
> 
> # From https://github.com/aaugustin/websockets/blob/main/example/echo.py
> 
> import asyncio
> import websockets
> import os
> 
> LOCAL_WS_SERVER_PORT = os.environ.get('LOCAL_WS_SERVER_PORT', '8765')
> 
> 
> async def echo(websocket, path):
>     async for message in websocket:
>         await websocket.send(XXX)
> 
> 
> async def main():
>     async with websockets.serve(echo, "localhost", LOCAL_WS_SERVER_PORT):
>         await asyncio.Future()  # run forever
> 
> asyncio.run(main())
> ```
> 
> **Client code** (lightly modified from Long-Lived Connection code)
> 
> ```
> import websocket
> import _thread
> import time
> import rel
> 
> def on_message(ws, message):
>     print(message)
> 
> def on_error(ws, error):
>     print(error)
> 
> def on_close(ws, close_status_code, close_msg):
>     print("### closed ###")
>     print("close_status_code: " + str(close_status_code))
>     print("close_msg: " + close_msg)
> 
> def on_open(ws):
>     print("Opened connection")
>     ws.send("hello")
> 
> if __name__ == "__main__":
>     websocket.enableTrace(True)
>     ws = websocket.WebSocketApp("ws://127.0.0.1:8765",
>                               on_open=on_open,
>                               on_message=on_message,
>                               on_error=on_error,
>                               on_close=on_close)
> 
>     ws.run_forever(dispatcher=rel)  # Set dispatcher to automatic reconnection
>     rel.signal(2, rel.abort)  # Keyboard Interrupt
>     rel.dispatch()
> ```

pouya817:
Hello dear,
now I have an issue with this part ['rel'].
my problem is my code connects to a socket successfully but my problem showed the during connection.
error code 1001 CloudFlare WebSocket proxy restarting,
I should use a thread in my code and the 'rel' does not support multithread activity and raises an error.
what can I do to solve this problem ( I mean disconnected from server unwanted )

`    def start(self):
        self.ws = websocket.WebSocketApp(self.subscribe_url,
                                         on_message=self.on_message, on_error=self.on_error, on_close=self.on_close)
        self.ws.on_open = self.on_open
        self.websocket_thread = threading.Thread(target=self.ws.run_forever, kwargs={'reconnect': True})
        self.websocket_thread.start()

    def stop(self):
        if self.ws:
            self.ws.close()
        if self.websocket_thread:
            self.websocket_thread.join()
        logger.info({'msg': f' count of datahub thread: {threading.active_count()}'})

    def on_close(self, ws, close_status_code, close_msg):
        sub_str = json.dumps(self.channel)
        logger.info({'msg': f'Unsubscribed {sub_str}',
                     'close_status_code': close_status_code, 'close_msg': close_msg})

    def on_message(self, ws, message):
        compressed_data = gzip.GzipFile(fileobj=io.BytesIO(message), mode='rb')
        decompressed_data = compressed_data.read()
        utf8_data = decompressed_data.decode('utf-8')
        if utf8_data == 'Ping':
            ws.send('Pong')
        elif json.loads(utf8_data).get('code') is not None:
            print(json.loads(utf8_data))
            self.callback(msg=json.loads(utf8_data))
    def on_open(self, ws):
        sub_str = json.dumps(self.channel)
        ws.send(sub_str)
        msg = f'Start subscribing to {sub_str}'
        logger.info({'msg': msg})

    def on_error(self, ws, error):
        sub_str = json.dumps(self.channel)
        msg = f'Websocket error: {error}, {sub_str}'
        logger.error({'msg': msg})`

bubbleboy14:
Regarding the threading/rel conflict, @pouya817, what's the error?

The reconnect issue is with your run_forever() kwargs. In particular, you currently have True for reconnect, but that should be an integer - try 1.

If you still have problems, try out this branch I've been working on:

https://github.com/websocket-client/websocket-client/tree/asyncpong

It should work with rel 0.4.9.11, but feel free to try the current working branch:

https://github.com/bubbleboy14/registeredeventlistener/tree/bwtweax

Hope it helps, LMK how it goes!

PR #854, opened 2022-08-26
fix for stack growth on reconnect

bubbleboy14:
previously, using the default (synchronous) dispatcher, reconnecting would cause the stack to grow.

here's the fix.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -39,6 +39,10 @@ class DispatcherBase:
         self.app = app
         self.ping_timeout = ping_timeout
 
+    def timeout(self, seconds, callback):
+        time.sleep(seconds)
+        callback()
+
 
 class Dispatcher(DispatcherBase):
     """
@@ -55,10 +59,6 @@ class Dispatcher(DispatcherBase):
                     break
             check_callback()
             sel.close()
-
-    def timeout(self, seconds, callback):
-        time.sleep(seconds)
-        callback()
 
 
 class SSLDispatcher(DispatcherBase):
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -30,15 +30,19 @@ limitations under the License.
 
 __all__ = ["WebSocketApp"]
 
-
-class Dispatcher:
+class DispatcherBase:
     """
-    Dispatcher
+    DispatcherBase
     """
     def __init__(self, app, ping_timeout):
         self.app = app
         self.ping_timeout = ping_timeout
 
+
+class Dispatcher(DispatcherBase):
+    """
+    Dispatcher
+    """
     def read(self, sock, read_callback, check_callback):
         while self.app.keep_running:
             sel = selectors.DefaultSelector()
@@ -51,15 +55,15 @@ class Dispatcher:
             check_callback()
             sel.close()
 
+    def timeout(self, seconds, callback):
+        time.sleep(seconds)
+        callback()
 
-class SSLDispatcher:
+
+class SSLDispatcher(DispatcherBase):
     """
     SSLDispatcher
     """
-    def __init__(self, app, ping_timeout):
-        self.app = app
-        self.ping_timeout = ping_timeout
-
     def read(self, sock, read_callback, check_callback):
         while self.app.keep_running:
             r = self.select()
@@ -95,6 +99,9 @@ class WrappedDispatcher:
     def read(self, sock, read_callback, check_callback):
         self.dispatcher.read(sock, read_callback)
         self.ping_timeout and self.dispatcher.timeout(self.ping_timeout, check_callback)
+
+    def timeout(self, seconds, callback):
+        self.dispatcher.timeout(seconds, callback)
 
 
 class WebSocketApp:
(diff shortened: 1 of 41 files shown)


websocket/_app.py
```diff
@@ -96,14 +105,18 @@ def __init__(self, app, ping_timeout, dispatcher):
...
         self.dispatcher = dispatcher
+        dispatcher.signal(2, dispatcher.abort) # keyboard interrupt
 
     def read(self, sock, read_callback, check_callback):
         self.dispatcher.read(sock, read_callback)
-        self.ping_timeout and self.dispatcher.timeout(self.ping_timeout, check_callback)
+        self.ping_timeout and self.timeout(self.ping_timeout, check_callback)
 
     def timeout(self, seconds, callback):
         self.dispatcher.timeout(seconds, callback)
```
engn33r, line 115: Should this be `self.timeout(seconds, callback)` instead of `self.dispatcher.timeout(seconds, callback)`?

websocket/_app.py
```diff
@@ -331,7 +344,7 @@ def teardown(close_frame=None):
             # Finally call the callback AFTER all teardown is complete
             self._callback(self.on_close, close_status_code, close_reason)
 
-        def setSock():
+        def setSock(reconnecting=False):
```
engn33r, line 347: What is this `reconnecting` variable useful for? I see it in line 370 but I don't think I'm clear on how its use on that line impacts the reconnect process. This variable name is very close to `reconnect`, so a different name that better differentiates would be better.

engn33r:
Added a couple comments. Would be good if you can fix the lint errors from CI (line 52 and 108)

websocket/_app.py
bubbleboy14: Yes - it's defined on line 114

websocket/_app.py
bubbleboy14: That's so that on line 370 it doesn't call handleDisconnect() again if it's already "reconnecting". That's actually the heart of this pull request - in synchronous (default) mode, the stack grows if handleDisconnect() is called (by way of a couple intermediate functions) recursively. Thus, when DispatcherBase.reconnect() (line 51) calls reconnector() (which is setSock()), it passes reconnecting=True, so that if the connection fails it exits setSock(), returns to DispatcherBase.reconnect() (popping the setSock() frame off the stack), and tries again after the wait interval.

bubbleboy14: And of course I'm open to changing "reconnecting" if there's a better name for that kwarg :)

bubbleboy14:
Ok we're all set on the lint errors

engn33r:
I wasn't sure why the CI is failing, but I investigated by duplicating the CI steps found in .github/workflows/build.yml. The changes to `run_forever` are breaking some unit tests. It's possible the tests were design to work with a run_forever function that doesn't reconnect like it does now and need substantial changes. I will plan to merge this PR after I fix the broken tests in the next week or two unless there is a need to merge this change sooner. If @bubbleboy14 or someone else wants to investigate the unit tests sooner, you're welcome to.

aantn:
Does the stack grow in version 1.3.3 as well or only 1.4.0?

ollo69:
Hi, I don't know about stack grow, but for sure 1.4.0 is introducing a bug when used with SSLDispatcher that is causing error `AttributeError: 'SSLDispatcher' object has no attribute 'timeout'` during disconnect. This was also fixed in PR #855 that was closed because was containing same code of this PR. I think this should be fixed with priority, doesn't it?

aantn:
We've rolled back to version 1.3.3 - the upgrade to version 1.4.0 was broke something for us. Our server might be responsible too, but the client behaviour in 1.3.3 didn't cause any trouble.

websocket/tests/test_app.py
```diff
@@ -6,6 +6,7 @@
 import websocket as ws
 import ssl
 import unittest
+import rel
```
bubbleboy14, line 9: This is to fix the testRunForeverDispatcher() test -- the 'dispatcher="Dispatcher"' previously on line 101 isn't correct.

I'm not sure if lines 102-103 will actually make the test work, I don't know how to add rel to the test dependencies.

Do these tests work in the master branch? As far as I can tell, testRunForeverDispatcher() at least doesn't seem like it could have.

bubbleboy14:
Hello all. @engn33r, all the tests are passing now, with the exception of the Codecov tests (which indicate decreased coverage), which makes sense, because there aren't tests yet for the new code.

I disabled the testRunForeverDispatcher() test because it can only work with rel, but I don't know how to add rel to the test dependencies. Also, I don't think this test is working in master.

I propose we merge this in and call it 1.4.1 for @aantn and @ollo69 and the other ha-samsungtv-smart people.

engn33r:
Nice work @bubbleboy14! I will merge and create a new release. With all the effort you've put in lately, send me an email at websocket.client.email [at] gmail.com if you're interested in joining as a co-maintainer.

bubbleboy14:
> Actually found that the error is coming from [THIS](https://github.com/websocket-client/websocket-client/blob/3baacdafe246c4eb93b41850bb29bda7b73c5fc3/websocket/_app.py#L94) line

Funny thread about sock.pending(): https://bugs.python.org/issue21430

These people in 2014 were saying "this function should not be used" - maybe it got removed in a recent version of Python.

Feel up to patching this @maksimu?

bubbleboy14:
@maksimu do you have a concise test case for this? My wss tests aren't failing.

Issue #858, opened 2022-09-09
ping/pong timeout recursion


adamwojt:
Discovered (thankfully in dev env) issue with ping/pong timeout when running with `run_forever`. Chain traceback, with multiple (10,000+) ping/pong timeout errors. At the end, our logger raised recursion error as it couldn't keep up. We are running with proxy and these parameters:
``` python
ping_timeout: int = 4
ping_interval: int = 5
``` 

``` log
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/websocket/_app.py", line 354, in setSock
    dispatcher.read(self.sock.sock, read, check)
  File "/usr/local/lib/python3.9/site-packages/websocket/_app.py", line 56, in read
    check_callback()
  File "/usr/local/lib/python3.9/site-packages/websocket/_app.py", line 397, in check
    raise WebSocketTimeoutException("ping/pong timed out")
websocket._exceptions.WebSocketTimeoutException: ping/pong timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/websocket/_app.py", line 354, in setSock
    dispatcher.read(self.sock.sock, read, check)
  File "/usr/local/lib/python3.9/site-packages/websocket/_app.py", line 56, in read
    check_callback()
  File "/usr/local/lib/python3.9/site-packages/websocket/_app.py", line 397, in check
    raise WebSocketTimeoutException("ping/pong timed out")
websocket._exceptions.WebSocketTimeoutException: ping/pong timed out

During handling of the above exception, another exception occurred:

....
```

bubbleboy14:
Hello @adamwojt . Could you please provide a minimal test case (client and server if possible) that reproduces the issue?

adamwojt:
@bubbleboy14 

Sure here is how i was able to reproduce it locally with `1.4.0` (same as in our env at the time of issue occurring):

**Install**:
1. `pip install websocket-client==1.4.0`
2. `pip install websockets`

**client.py**
``` python
import websocket
import logging
logging.basicConfig()

websocket_url = "ws://localhost:8765"
ping_timeout = 1
ping_interval = 2

def on_message(ws, data):
    print(data)

def on_error(ws, error):
    logging.exception("error")

def on_open(ws):
    print("Connection opened")

ws = websocket.WebSocketApp(
    websocket_url,
    on_message=on_message,
    on_error=on_error,
    on_open=on_open,
)

[...]

PR #859, opened 2022-09-13
WebSocketApp._send_ping(): omit break on ping failure

bubbleboy14:
omit break that causes pings to halt after failure (even post-reconnect).

may fix:

https://github.com/websocket-client/websocket-client/issues/858


engn33r:
This edit is included as one of the changes made in PR #862, so I will close this PR and merge the other

PR #864, opened 2022-10-03
Possible* Redundancy In Key Generation

benlazzero:
Was digging through the code base the past few days as it's been an excellent aid in learning about Websockets and TCP. Noticed _create_sec_websocket_key was getting called twice per ws.connect method. I may be missing something, but thought I would bring it to your attention in case it's an artifact or something. Best Regards!


engn33r:
Great catch, thanks for the PR!

PR #869, opened 2022-10-21
feat(*): return opcodes

r-putra-rp:


Coins2payroll (approved):

engn33r:
I don't understand some of the reasoning behind this PR so I am closing it. The changes that confuse me are:
- It would help if you could provide an explanation in the PR for why you made these changes and in what scenarios these changes are helpful
- The logic added to `validate()` is in the branch `if self.opcode == ABNF.OPCODE_CLOSE` so it will always return the OPCODE_CLOSE opcode
- Return values are added to functions without the function definition changing to reflect the added return values

Issue #857, opened 2022-09-07
'socket' object has no attribute 'pending'


j-linds:
Hi,

Since version 1.4.0 we have been experiencing an issue using `WebSocketApp` and obtaining this error:
```
'socket' object has no attribute 'pending'
```

Downgrading to an older version, such as `pip install websocket-client==1.3.3`, resolves the problem.

Example code:
```
#!/usr/bin/env python3
import websocket
import ssl
import json

message_to_send = {"key": "value"}

def on_message(ws, message):
    print("response: " + str(message))
    ws.close()

def on_error(ws, error):
    print("error: " + str(error))

def on_close(ws, *args, **kwargs):
    print("Closed")

def on_open(ws):
    msg = json.dumps(message_to_send)
    print(msg)
    ws.send(msg)

ws = websocket.WebSocketApp(
    "ws://localhost:8089/",
    on_message=on_message,
    on_error=on_error,
    on_close=on_close,
)
ws.on_open = on_open
ws.run_forever(sslopt={"cert_reqs": ssl.CERT_NONE})
```

Some example debugging information:
```
--- request header ---
GET / HTTP/1.1
Upgrade: websocket
Host: localhost:8089
Origin: http://localhost:8089
Sec-WebSocket-Key:
Sec-WebSocket-Version: 13
Connection: Upgrade

-----------------------
--- response header ---
HTTP/1.1 101 Switching Protocols
Connection: Upgrade
Sec-WebSocket-Accept:
Server: server
Upgrade: websocket
-----------------------

websocket connected
'socket' object has no attribute 'pending' - goodbye
error: 'socket' object has no attribute 'pending'
```

Have searched online/FAQ etc and cannot see anything related to the issue.

Many thanks for the help.

engn33r:
The same issue was reported in this comment and follow-up comments: https://github.com/websocket-client/websocket-client/pull/854#issuecomment-1238751577

I don't have more information other than what is mentioned there at the moment but will update with details when I do.

bubbleboy14:
Hello @j-linds , shucks sorry to hear. I ran your test against a wss server I run out in the wild, and I didn't notice any problems, or see the error you encountered.

Could you please provide your server code as well?

j-linds:
@bubbleboy14 This specific server code I cannot share, but I'll try to create a reproducible example and report back.

cyr123:
[...]

PR #875, opened 2022-11-01
create_dispatcher() w/ is_ssl determined by url instead of sslopt

bubbleboy14:
see: https://github.com/websocket-client/websocket-client/issues/857

### DIFF
--- a/websocket/tests/test_app.py
+++ b/websocket/tests/test_app.py
@@ -191,7 +191,7 @@ class WebSocketAppTest(unittest.TestCase):
         """ Test WebSocketApp binary opcode
         """
         # The lack of wss:// in the URL below is on purpose
-        app = ws.WebSocketApp('streaming.vn.teslamotors.com/streaming/')
+        app = ws.WebSocketApp('wss://streaming.vn.teslamotors.com/streaming/')
         app.run_forever(ping_interval=2, ping_timeout=1, ping_payload="Ping payload")
 
     @unittest.skipUnless(TEST_WITH_INTERNET, "Internet-requiring tests are disabled")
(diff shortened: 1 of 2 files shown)


engn33r:
The failing doctest tests are because websocket.events doesn't resolve, which is unrelated to the changes made to the code

engn33r:
Can you use `parse_url` instead? It returns a bool `is_secure` value that is used elsewhere for determining whether SSL is used and uses urlparse underneath which should handle more edge cases in the URL input
https://github.com/websocket-client/websocket-client/blob/664b0898b07cd054f8d0fb08c73a24dbb3095d5b/websocket/_http.py#L76

bubbleboy14:
> Can you use `parse_url` instead? It returns a bool `is_secure` value that is used elsewhere for determining whether SSL is used and uses urlparse underneath which should handle more edge cases in the URL input
> 
> https://github.com/websocket-client/websocket-client/blob/664b0898b07cd054f8d0fb08c73a24dbb3095d5b/websocket/_http.py#L76

Good call @engn33r , PR updated

engn33r:
Thanks for the PR @bubbleboy14! :+1:

websocket/tests/test_app.py
```diff
@@ -191,7 +191,7 @@ def testOpcodeBinary(self):
         """ Test WebSocketApp binary opcode
         """
         # The lack of wss:// in the URL below is on purpose
```
garethsb, line 193: This comment needs changing to say the opposite, no? Including the scheme is now REQUIRED.

engn33r: good point, I'll remove the comment because it doesn't make sense. I had to comment out this test already because the URL no longer works

Issue #865, opened 2022-10-06
Feature Request: Suppress Logging/Warning


nijian84:
When the Websocket-client connects to a WS, it spits out a logging warning. These should be abled to be turned off, by some arg, or off by default, and enabled via the Trace function that exists already. Kinda causes my logs to be noisy.

Thanks!

engn33r:
I cannot tell which warnings you are describing but I expect setting [this to false or removing the `enableTrace` line will help](https://websocket-client.readthedocs.io/en/latest/examples.html#debug-and-logging-options).

NadavK:
I had the same issue with logs being outputted even with Trace set to False:
> 2022-09-18 10:22:01,571 - ERROR - websocket - [Errno -2] Name or service not known - reconnecting
> 2022-09-18 10:22:01,575 - INFO - websocket - reconnect() - retrying in 5 seconds [5 frames in stack]

Fixed with following code:
``` python
        websocket.enableTrace(True, level="FATAL")      # Setting "level" only works if Trace is enabled.
        websocket.enableTrace(False, level="FATAL")     # False still produces disconnect logs
```

engn33r:
This should be solved in v1.5.0 because PR #888 is merged in this release

PR #888, opened 2023-01-11
Reduce "websocket connected" log to INFO level

pR0Ps:
This doesn't seem like it should be a warning. A connection was requested and a connection was made so INFO seems like a more appropriate level for this message.

I believe this is the warning #865 was referring to.


engn33r:
Thanks, this makes sense

PR #898, opened 2023-02-27
fix: pass correct number of arguments to logging function

ozichert:
As the wrong number of arguments is passed to the logging function there occured an error.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -272,8 +272,8 @@ class WebSocketApp:
                 try:
                     _logging.debug("Sending ping")
                     self.sock.ping(self.ping_payload)
-                except Exception as ex:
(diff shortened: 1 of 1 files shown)


engn33r:
I would prefer to update the code to use str.format() instead of the older % operator, in case % gets deprecated in the future. Can you edit the PR?

ozichert:
@engn33r Just did this for the whole `_app.py` logging messages.

engn33r:
Thanks for the PR. At some point a PR should be made (by me or someone else) to switch the other instances of % to str.format() for consistency

engn33r:
Closing out this comment. I removed the modulo string formatting with % and replaced it with str.format() in https://github.com/websocket-client/websocket-client/commit/c6a445f2b98b4f80befcd5260da82d157e3fdabf

Issue #684, opened 2021-05-05
Project upgrade initiative 2021


engn33r:
Hello websocket users! :wave:

After taking over of maintainer of this project in January 2021, some good progress has been made, including:
- Reduced the 200+ open issues down to 25 open issues
- Created and substantially improving the project documentation
- Transitioned from Travis CI to GitHub Actions
- Released two new versions (0.58.0 and 0.59.0)
- Resolved all outstanding PRs

It's time to move to phase 2 of getting this project in order! Here's what I expect to get done:
- Solve the licensing of this library -> comment ASAP on issue #665 if you have opinions on this! I hope to decide on a license in mid-May
- Release v1.0.0 in mid-May after removing Python 2 code from library. 0.59.0 was released as the final main release supporting Python 2, and version 1.0.0 will be Python 3 only
- Fix readthedocs documentation builds so that it properly shows the comments embedded in the code
- Improve unit tests and increase code coverage to over 80%
- Get down to only 15 open issues
- Refactor some portions of the code to make it easier to understand
- Achieve full (or nearly full) flake8 compliance

For tracking purposes, issue #660 has my initial list of goals from soon after taking on the project maintainer role

instance01:
I just want to say - you are doing an absolutely amazing job at this. So Kudos to you!

bobcolner:
just dropped back here -amazed to see this project come back to life after raised about the status of the project 6 months ago.

adamwojt:
May I suggest of one more effort?
- Add optional static typing.

It would highly improve readability, type hints in IDEs, and make it easier to contribute to code. Later on when code-base would be fully typed, mypy could be used for static analysis and server as tests for making sure that correct arguments are being passed around. 

I remember that when I was creating PR https://github.com/websocket-client/websocket-client/pull/842/files it was quite hard to keep up of what goes where as untyped `**options` where passed around the code.

This could be achieved by including typing directly in code which is easier to maintain. The great example of fully typed library is [FastApi](https://github.com/tiangolo/fastapi/blob/master/fastapi/applications.py)

Another alternative is stub package similar to those here: https://github.com/python/typeshed/tree/master/stubs

I am happy to help in free time.

engn33r:
@adamwojt #749 is about static typing, we can discuss more there. Help would be very much appreciated, even if the PR only adds partial type hint support!

PR #908, opened 2023-05-14
Adding typehints to _abnf.py

### LINKED ISSUE #749: FR: add type hint
Since this project requires python >= 3.6, it would be great if we could add type hint support.

RoryGlenn:
Adding typehints to _abnf.py.
This will be one of several commits to help resolve #749 and #684.
I will try to commit one file at a time for easier debugging and reviewing.

### DIFF
--- a/websocket/_abnf.py
+++ b/websocket/_abnf.py
@@ -33,14 +33,14 @@ try:
     # Note that wsaccel is unmaintained.
     from wsaccel.xormask import XorMaskerSimple
 
-    def _mask(_m, _d):
+    def _mask(_m, _d) -> bytes:
         return XorMaskerSimple(_m).process(_d)
 
 except ImportError:
     # wsaccel is not available, use websocket-client _mask()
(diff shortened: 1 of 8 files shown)


engn33r:
Thank you, adding typehints is a very welcome addition! :+1:

The CI indicates these changes need some linting fixes and at least one test needs modification to run. I can take a look at what is needed if it would help, but don't have time this week.

engn33r:
I wonder if the current build error of `TypeError: 'type' object is not subscriptable` for the line `def _is_no_proxy_host(hostname: str, no_proxy: list[str]) -> bool:` is only an issue on Python 3.8?  I wasn't able to recreate the issue in a quick local test with a more recent Python version, I'll dig into this deeper later.

RoryGlenn:
Yes that is correct, python 3.8 does not allow nested type hints but 3.9 does. 
I'll submit a change this week

engn33r:
All CI is passing, thanks for the upgrades :+1:

PR #909, opened 2023-05-21
Sourcery-ai improvements

RoryGlenn:
Proposed Improvements.
These are changes suggested by sourcery-ai.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,13 @@
 ChangeLog
 ============
 
+- 1.5.2
+  - Add typehints (#908)
+  - Fix pytype errors (#906)
+  - Fix args passed to logging function (#898)
+  - Standardize PEP 3101 formatting (c6a445f)
+  - Add more verbose exception for unsuccessful handshake for #900 (f85ae1f)
+
 - 1.5.1
   - Fix logic bug that can cause disconnects (#893)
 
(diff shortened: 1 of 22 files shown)


engn33r:
My two cents is that some of these suggested improvements might not make the code easier to read. It is possible that some of the edits improve type handling or something else, specifically all the places where string concatenation is done.

I won't get around to reviewing all the changes here until later, but it might be better to break this PR up into multiple PRs to make it more manageable.

engn33r:
Thanks for the PR, even if AI did most of the work :smile: 

I manually reviewed all the changes. I applied most of the edits that I thought improved the code in commit 780584f2b1e664332559ec27a6b81041ccdc082e. Notably:
- Combined assignment and if statement logic with the "[walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions)" introduced in Python 3.8 to reduce the line count
- Simplified the logic and/or fixed minor bugs in a couple cases (for example,recv() in _core.py had an unnecessary ABNF.OPCODE_TEXT check)
- Cleanup and reorganization of if/else branching logic
- Used f-strings to replace other string formatting/concatenation approaches such as:
  1. the modulo operator %
  2. str.format()
  3. "+" for string concatenation

I did not replace the if/else statements with ternary statements because they can reduce readability in some locations.

PR #911, opened 2023-05-26
Relying on a header that might not be present, so added a header check

BrentWilkins:
I found a case where `_get_resp_headers` gets a `KeyError`.

Here are the `status`, `resp_headers`, and `status_message` that were obtained by `read_headers`:

```text
status = 404
resp_headers = {'date': 'Fri, 26 May 2023 22:34:07 GMT', 'server': 'WSGIServer/0.2 CPython/3.11.3', 'content-type': 'text/html', 'vary': 'Origin, Cookie', 'access-control-allow-origin': '*', 'connection': 'close'}
status_message = 'Not Found'
```

I looked at the HTTP spec:
> For response messages, whether or not an entity body is included with a message is dependent on both the request method and the response code. All responses to the HEAD request method must not include a body, even though the presence of entity header fields may lead one to believe they do. All 1xx (informational), 204 (no content), and 304 (not modified) responses must not include a body. All other responses must include an entity body or a Content-Length header field defined with a value of zero (0).

-- [spec](https://www.w3.org/Protocols/HTTP/1.0/draft-ietf-http-spec.html#Entity-Body)

I think that says that the server should have sent a Content-Length header, but maybe the code shouldn't crash?

After my change, I get this:
> websocket._exceptions.WebSocketBadStatusException: Handshake status 404 Not Found -+-+- {'date': 'Fri, 26 May 2023 22:53:52 GMT', 'server': 'WSGIServer/0.2 CPython/3.11.3', 'content-type': 'text/html', 'vary': 'Origin, Cookie', 'access-control-allow-origin': '*', 'connection': 'close'} -+-+- None

Maybe `response_body = ''` would be better, but here is something to start with.

### DIFF
--- a/websocket/_handshake.py
+++ b/websocket/_handshake.py
@@ -131,7 +131,7 @@ def _get_handshake_headers(resource, url, host, port, options):
     client_cookie = options.get("cookie", None)
 
     cookie = "; ".join(filter(None, [server_cookie, client_cookie]))
-    
+
     if cookie:
         headers.append("Cookie: {cookie}".format(cookie=cookie))
 
--- a/websocket/_handshake.py
+++ b/websocket/_handshake.py
@@ -144,7 +144,10 @@ def _get_handshake_headers(resource, url, host, port, options):
 def _get_resp_headers(sock, success_statuses=SUCCESS_STATUSES):
     status, resp_headers, status_message = read_headers(sock)
     if status not in success_statuses:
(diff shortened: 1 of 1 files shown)


engn33r:
CI is failing for Python 3.7 because [assignment expressions](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions), like `if cookie :=`, were added in Python 3.8. Python 3.7 EOL is in 1 month, so I can merge in 1 month after removing Python 3.7 support for websocket-client or the PR can be edited to support Python 3.7 syntax.

PR #912, opened 2023-05-31
Fix first round ping waiting 2 x interval.

SunFulong:
https://github.com/websocket-client/websocket-client/blob/7c39ddb91d6ed07974f3435d96e0d57a40435e54/websocket/_app.py#L270-L274

The first round of ping out will wait 2 times the interval, because there are 2 waits blocked, so, I added a variable to identify whether it is the first round.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -270,7 +270,9 @@ class WebSocketApp:
     def _send_ping(self) -> None:
         if self.stop_ping.wait(self.ping_interval):
             return
-        while not self.stop_ping.wait(self.ping_interval):
+        first_round = True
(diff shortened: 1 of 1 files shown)


engn33r:
Before merging this fix, can we test if the first if statement can be removed as an alternative? That would be a cleaner solution if it works.

engn33r:
I tested the alternate solution mentioned in my previous comment and it works based on my testing. So instead of the change in this PR, I suggest removing these two lines from `_send_ping()`:

```python
    if self.stop_ping.wait(self.ping_interval):
        return
```

Issue #917, opened 2023-06-12
WebSocketApp teardown() sometimes not called.


QuinnDamerell:
Hello! I have a Python extension that uses this lib heavily, and I found a bug where the WebSocket ping thread never gets shut down.

Due to this related bug #916, sometimes `Dispatcher.read()` can exit without calling `read_callback`, when the returned `r` value is None, as seen below.

`
            r = sel.select(self.ping_timeout)
            if r:
                if not read_callback():
                    break
            check_callback()
            sel.close()
`

If that happens, then nothing ever calls `teardown()` before `run_forever` exits. Normally the select would result in a read, which would call the read() callback, which would call teardown because keep_running is False:

`        def read():
            if not self.keep_running:
                return teardown()
       `
      
But since the read callback isn't called, the thread calling `run_forever` returns out of run_forever without calling teardown. Since there's no external interface for teardown(), there's then no way to call it, and thus nothing ever frees the send_ping thread.

I think there are two fixes here:

1) `_send_ping` should check the `keep_running` state as a fall back, so if it's failed to be stopped, it can still end.
2) The `run_forever` function should always ensure `teardown` is called before it exists, or the allocated resources will be leaked.

engn33r:
Excellent find, fixed in v1.6.0 by your PR #918

PR #918, opened 2023-06-12
Adding logic to ensure teardown is called before run_forever exits, otherwise resources like the ping thread are leaked.

QuinnDamerell:
Fixes issue #917, see my description there on how it's possible to return from `run_forever` without calling `teardown`. There might be other ways as well, so my solution here is a good way to ensure it's called.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -375,7 +375,7 @@ class WebSocketApp:
                 with the statusCode and reason from the provided frame.
             """
 
-            # teardown is called in many code paths to ensure resources are cleaned up and on_close is fired.
+            # teardown() is called in many code paths to ensure resources are cleaned up and on_close is fired.
             # To ensure the work is only done once, we use this bool and lock.
             with self.has_done_teardown_lock:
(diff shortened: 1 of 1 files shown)


engn33r:
Excellent work identifying the specific code path that causes this issue. I will request @bubbleboy14 to give a final comment before I merge this and create a new 1.6.0 minor release, but this PR looks like a very welcome upgrade.

I tested this updated code on the two problematic code examples where `on_close` was not getting triggered and this solves them :+1:
https://websocket-client.readthedocs.io/en/v1.5.2/threading.html#possible-issues-with-threading

bubbleboy14:
Great job @QuinnDamerell ! These look like excellent changes, and they pass my tests.

I say go for it!

QuinnDamerell:
@engn33r, no pressure, but do you have any timeline on when this change might happen? I really want to pull it into my PY app, because I have quite a few users who hit this.

I'm happy to help if you need any assistance in testing or whatnot to get it in!

Issue #923, opened 2023-06-21
In version 1.6.0 using dispatcher=rel immediately closes the socket


ptriller:
When I connect to a websocket that uses

```
ws.run_forever(dispatcher=rel, reconnect=5)
rel.dispatch()
```

and code that is identical to the example code in the README

with version 1.5.3 I get a successfull connect
```
++Sent raw: b'\x81\xc6`g!\x0c\x1bESi\x13\x08T~\x03\x02\x036@ErX! dS$"w.LG\x03|\x12\x08Bi\x13\x14\x036@EQi\x14\x02S>BK\x01.\x13\x12C|\x12\x08Bi\x13\x14\x036@EMc\x03\x06M>B\x1a'
++Sent decoded: fin=1 opcode=1 data=b'{"resource": "STAGE_DEV", "process": "peter2", "subprocess": "local2"}'
++Rcv raw: b'\x81\x10{"success":true}'
++Rcv decoded: fin=1 opcode=1 data=b'{"success":true}'
g
```

with version 1.6.0 I get an immediate disconnect:
```
++Sent raw: b'\x81\xc6\xc6\xc7\xfd9\xbd\xe5\x8f\\\xb5\xa8\x88K\xa5\xa2\xdf\x03\xe6\xe5\xaem\x87\x80\xb8f\x82\x82\xab\x1b\xea\xe7\xdfI\xb4\xa8\x9e\\\xb5\xb4\xdf\x03\xe6\xe5\x8d\\\xb2\xa2\x8f\x0b\xe4\xeb\xdd\x1b\xb5\xb2\x9fI\xb4\xa8\x9e\\\xb5\xb4\xdf\x03\xe6\xe5\x91V\xa5\xa6\x91\x0b\xe4\xba'
++Sent decoded: fin=1 opcode=1 data=b'{"resource": "STAGE_DEV", "process": "peter2", "subprocess": "local2"}'
++Sent raw: b'\x88\x82\x94\xb8.\xdf\x97P'
++Sent decoded: fin=1 opcode=8 data=b'\x03\xe8'
```
It works without the dispatch=rel

bubbleboy14:
@ptriller does https://github.com/websocket-client/websocket-client/pull/924 fix your issue?

@engn33r sorry, i should have tested the bidirectional case, which would have revealed this problem.

the pull request should fix the (bidirectional) custom dispatcher by preventing teardown() from getting immediately called in this case.

[...]

PR #924, opened 2023-06-21
DispatcherBase.reconnect(): raise KeyboardInterrupt to prevent gettin…

bubbleboy14:
…g stuck in reconnect loop. WebSocketApp.run_forever(): only teardown() in finally if not custom_dispatcher (avoids immediately closing connection); added except case that logs and calls teardown(), catching things like KeyboardInterrupt.

may fix: https://github.com/websocket-client/websocket-client/issues/923


engn33r:
Thanks for catching this. Adding KeyboardInterrupt is a good upgrade on its own, even if it does not solve any open issues. Github isn't completing the tests but they all pass when I test locally.

PR #929, opened 2023-07-10
Use https_proxy environment variable for secure connections

### LINKED ISSUE #927: Environment variable HTTP_PROXY is used for HTTPS connections
The problem occurred in an environment where a proxy server is to be used for HTTP connections but not for HTTPS connections.
In this case `http_proxy` is set in the environment but `https_proxy` is not.
The problematic code is here:

https://github.com/websocket-client/websocket-client/blob/bd506ad2e14749e1d31c07a1a6fca5644adb0ec4/websocket/_url.py#L161-L163

In my opinion, only `https_proxy` should be used for secure connections.

j-a-n:
Use https_proxy environment variable for secure connections and do not fallback to http_proxy.
Fixes #927

### DIFF
--- a/websocket/_url.py
+++ b/websocket/_url.py
@@ -137,7 +137,7 @@ def get_proxy_info(
         Websocket server name.
     is_secure: bool
         Is the connection secure? (wss) looks for "https_proxy" in env
-        before falling back to "http_proxy"
+        instead of "http_proxy"
     proxy_host: str
         http proxy host name.
     proxy_port: str or int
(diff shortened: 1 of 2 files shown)


engn33r:
I agree with the update. Although the user is expected to set `is_secure` to true when `https_proxy` is set, the updated code makes it more clear and foolproof that `is_secure` controls whether `http_proxy` or `https_proxy` is chosen. This avoids the case of falling back to `http_proxy` if `https_proxy` is not set when the user specifically sets `is_secure` to true.

Thanks for adding test coverage :+1:

j-a-n:
Thanks for merging 👍

PR #932, opened 2023-07-22
Callable `header` argument on `WebSocketApp`

figaro-smartotum:
This pull request add the possibility to pass a "callable headers" argument to the WebSocketApp  object.

In this way it is possible to update the headers value just-before the connection attempt.

I needed to do this modification since I have an authentication scheme that requires an header containing a valid timestamp, so the headers needs to be updated before trying the connection in the background thread in order to keep the time in-sync.

Example:

```python
import time
import websocket 

def get_header() -> dict:
    """
    This function is called before every connection attempt
    :return: a dict containing the headers sent to the websocket
    """

    return {"X-My-Timestamp": str(time.time())}

ws = websocket.WebSocketApp("<my-endpoint>", header=get_header, on_message=lambda ws, msg: print(msg))
...

```

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -139,7 +139,7 @@ class WebSocketApp:
     Higher level of APIs are provided. The interface is like JavaScript WebSocket object.
     """
 
-    def __init__(self, url: str, header: list or dict = None,
+    def __init__(self, url: str, header: list or dict or Callable = None,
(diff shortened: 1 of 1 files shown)


engn33r:
A simple change and I can see reasonable use cases for it. Thanks for the PR

PR #945, opened 2023-08-27
Fix type annotations in the with str | bytes

1st:
I've faced an issue with Mypy that it complains about passing string to the ws.send() method. And found that you use bytes or str instead of the Python built-in pipe to be `str | bytes`.

Checked with Mypy and it works fine now.

### DIFF
--- a/websocket/_core.py
+++ b/websocket/_core.py
@@ -324,7 +324,7 @@ class WebSocket:
         """
         return self.send(payload, ABNF.OPCODE_BINARY)
 
-    def ping(self, payload: str or bytes = ""):
+    def ping(self, payload: str | bytes = ""):
         """
         Send ping data.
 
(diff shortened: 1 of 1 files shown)


engn33r:
I reverted this PR with a3f4c3a7501bb134e9a37d678b490feb188f0c00 because it is breaking Python 3.8 and 3.9 CI and compatibility. The `|` union operator was added in Python 3.10 in [PEP 604](https://peps.python.org/pep-0604/).

1st:
Yes. It's modern syntax. If you need an old syntax then create a branch for supporting old version of Python. Or you want to have a single version version that covers old and new versions?

engn33r:
There is only a single version of the library to support all versions of python that are not EOL. The supported versions right now are 3.8+, listed in [setup.py](https://github.com/websocket-client/websocket-client/blob/master/setup.py)

Issue #948, opened 2023-09-08
_abnf.py/frame_buffer constructor is annotated incorrectly


jolaf:
I observe the following crash under `typeguard`:
```
Traceback (most recent call last):
  File "test.py", line 750, in test
    webSocket = create_connection(WEBSOCKET_URL, sslopt = {"ca_certs": "/etc/ssl/certs/rootCA.crt"})
  File "/home/buildmaster/.local/lib/python3.10/site-packages/websocket/_core.py", line 603, in create_connection
    websock = class_(sockopt=sockopt, sslopt=sslopt,
  File "/home/buildmaster/.local/lib/python3.10/site-packages/websocket/_core.py", line 94, in __init__
    self.frame_buffer = frame_buffer(self._recv, skip_utf8_validation)
  File "/home/buildmaster/.local/lib/python3.10/site-packages/websocket/_abnf.py", line 276, in __init__
    def __init__(self, recv_fn: int, skip_utf8_validation: bool) -> None:
  File "/home/buildmaster/.local/lib/python3.10/site-packages/typeguard/_functions.py", line 138, in check_argument_types
    check_type_internal(value, annotation, memo)
  File "/home/buildmaster/.local/lib/python3.10/site-packages/typeguard/_checkers.py", line 766, in check_type_internal
    raise TypeCheckError(f"is not an instance of {qualified_name(origin_type)}")
typeguard.TypeCheckError: argument "recv_fn" (method) is not an instance of int
```
The problem is in `_abnf.py` line 276:
`def __init__(self, recv_fn: int, skip_utf8_validation: bool) -> None:`
As far as I could understand, `recv_fn` is expected to be a function returning `int`, not just `int` as specified.
Probably the correct annotation would be `Callable[[int], int]`.

jolaf:
If the project has a test set, I recommend running it under `typeguard`, it would expose most incorrect annotations.

engn33r:
This should be fixed by #949

The first goal to improve typing is to add mypy to CI and fix mypy errors. Then typeguard in the tests can come after

PR #949, opened 2023-09-09
Broken typing annotations fixed

### LINKED ISSUE #934: Typing annotations should use `|` instead of `or`
See, for example, `websocket/_socket.py/setdefaulttimeout()` and `getdefaulttimeout()` headers.
`int or float` is resolved as simply `int` which is just incorrect.

Also, `getdefaulttimeout()` sometimes returns `None` which is not reflected in the return type annotation.
It should be `def getdefaulttimeout() -> int | float | None:`.

Also, `websocket/_core.py/settimeout()` sometimes accepts `None` argument which is not reflected in argument type annotation.

Also, `websocket/_core.py/get_proxy_info()` header should be: `def get_proxy_info(hostname: str, is_secure: bool, proxy_host: str | None = None, proxy_port: int = 0, proxy_auth: tuple | None = None, no_proxy: list | None = None, proxy_type: str = 'http') -> tuple:`

Also, `websocket/_core.py/_is_no_proxy_host()` header should be: `def _is_no_proxy_host(hostname: str, no_proxy: list | None) -> bool:`, as `no_proxy` sometimes comes `None`.

Also, `websocket/_socket.py/send()` headers should be: `def send(sock: socket.socket, data: bytes | str) -> int:` and `def send(self, payload: bytes | str, opcode: int = ABNF.OPCODE_TEXT) -> int:` as they accept both `bytes` and `str`.

Also, `websocket/_cookiejar.py/add()` header should be: `def add(self, set_cookie: str | None) -> None:` as it also accepts `None`.

Also, `websocket/_abnf.py/create_frame()` header should be: `def create_frame(data: bytes | str, opcode: int, fin: int = 1) -> 'ABNF':` as it accepts both `bytes` and `str`.

Thanks to `typeguard` for catching these.

I would recommend using `typeguard` when running tests, it would catch bugs like this.

jolaf:
Running an app using `websocket-client` under `typeguard` produced a number of crashes due to typing annotations in `websocket-client` being broken and not compliant to the real data types. This PR fixes those broken annotations.

Fixes #934 and #948.

### DIFF
--- a/websocket/_abnf.py
+++ b/websocket/_abnf.py
@@ -1,13 +1,13 @@
-from collections.abc import Callable
-
 import array
 import os
 import struct
 import sys
 
+from threading import Lock
+from typing import Callable, Union
+
 from ._exceptions import *
 from ._utils import validate_utf8
-from threading import Lock
 
 """
 _abnf.py
@@ -134,7 +134,7 @@ class ABNF:
     LENGTH_63 = 1 << 63
 
     def __init__(self, fin: int = 0, rsv1: int = 0, rsv2: int = 0, rsv3: int = 0,
-                 opcode: int = OPCODE_TEXT, mask: int = 1, data: str | bytes = "") -> None:
+                 opcode: int = OPCODE_TEXT, mask: int = 1, data: Union[str, bytes] = "") -> None:
         """
         Constructor for ABNF. Please check RFC for arguments.
         """
@@ -189,7 +189,7 @@ class ABNF:
             + " data=" + str(self.data)
 
     @staticmethod
-    def create_frame(data: bytes | str, opcode: int, fin: int = 1) -> 'ABNF':
+    def create_frame(data: Union[bytes, str], opcode: int, fin: int = 1) -> 'ABNF':
         """
         Create frame to send text, binary and other data.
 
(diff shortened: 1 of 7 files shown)


engn33r:
Using `|` instead of `or` is not supported in Python 3.8 or 3.9 because [PEP 604](https://peps.python.org/pep-0604/) was introduced in Python 3.10. This is the same reason PR #945 was not included. This is why the CI tests are failing.

Currently a single release of websocket-client is intended to support all Python versions that are not EOL. If you'd suggest a change in strategy to support multiple versions simultaneously, I'm open to ideas.

jolaf:
Concerning failing tests. The errors like
`TypeError: unsupported operand type(s) for |: 'type' and 'type'`
are due to the fact that the test are being run on Python 3.8, while syntax like `int | str` is valid since Python 3.10.
I suggest switching test run to a newer Python.

This syntax appeared due to fixes I've made in this PR, it was `int or str`, and I've changed it to `int | str`. The former is just invalid.

If you want typing annotations valid for Python 3.8, you should write `Union[str, int]` instead, which is much more cumbersome.

engn33r:
I looked into this more, you're right. Even mypy outputs an error of "Invalid type comment or annotation" on the lines using `or`.

In order to maintain backwards compatibility with 3.8, would you be willing to replace the usage of `|` with `Union[]` or a similar change to allow the existing CI to pass? I can work on editting the changes in this PR if you prefer.

jolaf:
I've fixed the annotations, please run the checks.

engn33r:
Thank you for the helpful upgrade!

PR #950, opened 2023-09-16
Create wsdupm.py

aleksejaku:


engn33r:
I don't understand what this PR intended to do, but I do not see any need to include this

PR #958, opened 2023-12-04
Update _app import for type checkers

vchrisb:
Pylint does comlains when importing WebSocketApp from websocket  with:

> "WebSocketApp" is not exported from module "websocket"
>  Import from "websocket._app" instead Pylance([reportPrivateImportUsage](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportPrivateImportUsage))

To make type checkers happy, I updated the import..


engn33r:
Thanks :+1:

PR #961, opened 2023-12-15
WebSocketApp.run_forever(), read(): check for ssl.SSLEOFError; on exc…

bubbleboy14:
…eption, if custom_dispatcher, pass bool(reconnect) to handleDisconnect()

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -1,3 +1,4 @@
+import ssl
 import inspect
 import selectors
 import socket
@@ -516,9 +517,10 @@ class WebSocketApp:
             except (
                 WebSocketConnectionClosedException,
                 KeyboardInterrupt,
+                ssl.SSLEOFError,
             ) as e:
                 if custom_dispatcher:
-                    return handleDisconnect(e)
+                    return handleDisconnect(e, bool(reconnect))
                 else:
                     raise e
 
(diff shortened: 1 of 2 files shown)


websocket/_app.py
```diff
@@ -1,3 +1,4 @@
+import ssl
```
engn33r, line 1: Change this to `from ._ssl_compat import *` like the other files. _ssl_compat.py exists for the edge case where SSL can't be imported

engn33r:
The change to specify the reconnecting bool value to `handleDisconnect` makes sense to me. I'd suggest a different approach for `import ssl` or the EOF error, though I'm not sure how exactly to replace it. It may need edits to _ssl_compat.py.

This finding shows that better test coverage of the different branches would be a good improvement to work on in the future :+1:

bubbleboy14:
Good call @engn33r I made the suggested changes.

Just realized I did "from ._ssl_compat import SSLEOFError" instead of the recommended "from ._ssl_compat import *". I kind of like the single import more, but on the other hand maybe we should be checking for the other exceptions as well.

In either case, this feels like an improvement for now, and certainly addresses the edge case I encountered.

engn33r:
Great, thanks for the fix!

Issue #971, opened 2024-02-25
Multiple on_open events


NadavK:
Client receives an occasional second `on_open` callback, even though I did not actively re-open a connection, not did I receive an `on_close` or `on_error` callback.
The `on_close` and `on_error` callbacks _do_ get fired at other times, so the problem is not a mistake in my callback function/log.
Using v1.7.0.

client-side log:
22:13:02 - ws opened (<websocket._app.WebSocketApp object at 0xffff98662250>)
22:13:02 - ws sending: {msg 1...}
22:13:20 - ws sending: {msg 2...}
22:19:52 - ws on_ws_error Connection to remote host was lost. (<websocket._app.WebSocketApp object at 0xffff98662250>)
22:21:04 - ws opened (<websocket._app.WebSocketApp object at 0xffff98662250>)
22:38:52 - ws opened (<websocket._app.WebSocketApp object at 0xffff98662250>)
23:07:47 - ws closed: close_status_code: None, close_msg: "None" (<websocket._app.WebSocketApp object at 0xffff98662250>)

server side log (Django channels):
22:13:02 - connected _(correlates to client `on_open`)_
22:19:52 - disconnect _(correlates to client `on_error`)_
22:21:04 - connected _(correlates to client `on_open`)_
22:38:52 - connected _(correlates to client second `on_open`)_
22:39:15 - disconnect _(no client correlation)_
23:07:47 - disconnect _(correlates to client `on_close`)_

This behavior affects my server-side client-status, since the disconnect at 22:39:15 sets the server-side status to "not-connected", even though there is still an active connection and data sent by the server is still received by the client.
I can solve it by using a reference count on the server, but in any case the client behavior seems odd...

bubbleboy14:
Looks like this line should be conditionalized: https://github.com/websocket-client/websocket-client/blob/a394b46c560cb78672f95bbb3836305e8511d1b9/websocket/_app.py#L500

What do you think @engn33r ? For starters, we could just set a flag the _first_ time _callback(on_open) gets called from setSock(), and not call it subsequent times.

Also, though, since there are probably _some_ applications that would want on_open() called more than once (presumably here on reconnects), I think it might be nice to add a kwarg flag to run_forever() to specify whether or not on_open should only be called the first time.

Or (and this might be cleaner) we could add an on_reconnect() callback? And setSock() could just call on_open() the first time and on_reconnect() from then on. I think I like that approach most.

Thoughts?

NadavK:
[...]

PR #972, opened 2024-03-01
on_reconnect callback

bubbleboy14:
Ticket 971 (https://github.com/websocket-client/websocket-client/issues/971) discusses an ambiguity in our callback system between the connection opening for the first time and subsequent automatic reconnects.

To address this, this PR adds support for an on_reconnect callback. The on_reconnect callback (instead of the on_open callback) is called from setSock() on a reconnect. If no on_reconnect callback is provided, the previous behavior (of calling on_open on a reconnect) is preserved.

Ticket 974 (https://github.com/websocket-client/websocket-client/issues/974) involves the somewhat related case of WebSocketApp reuse after the server _intentionally_ disconnects. @GooVincent identified the solution, setting has_done_teardown to False.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -503,7 +503,7 @@ class WebSocketApp:
                 if self.ping_interval:
                     self._start_ping_thread()
 
-                if reconnecting:
+                if reconnecting and self.on_reconnect:
                     self._callback(self.on_reconnect)
                 else:
                     self._callback(self.on_open)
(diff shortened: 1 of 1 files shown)


engn33r:
This looks reasonable to me. Adding the new `on_reconnect` callback gives the user more flexibility without any compatibility issues. Setting has_done_teardown to false in run_forever makes sense to fix this case, it's a nice catch.

Issue #976, opened 2024-03-19
reconnection fails when server returns op_code = ABNF.OPCODE_CLOSE


polypauldeep:
Hello, I'm an active user of this library, using it to gather data via websocket.

I noticed the reconnect option available when employing the `run_forever` method. However, I've encountered a situation where if the server issues an `OPCODE_CLOSE` (8) message, the `run_forever` execution concludes. This occurs because the control flow is directed to:

https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L538-L539

resulting in the teardown of the connection.

Is this behavior by design? My understanding was that the connection should attempt a reconnect using the reconnect parameter, irrespective of the server's message. It appears that this situation arises because it's not captured by `WebSocketConnectionClosedException`, leading to `keep_running` returning false prior to reaching:

https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L619-L623

polypauldeep:
I've made an adjustment to one line in order to address the issue mentioned above.
https://github.com/polypauldeep/websocket-client/commit/360c33fcbb26e42148c302f1cf6cb669ea1179bf

https://github.com/polypauldeep/websocket-client/blob/360c33fcbb26e42148c302f1cf6cb669ea1179bf/websocket/_app.py#L538-L539

Let's have a discussion before I proceed with creating a pull request for this change.

engn33r:
This looks similar to PR #972, which was merged a couple weeks ago but I forgot to push a new version since then. Can you pull the latest code from master and check if that solves your problem? I can release a new version after you confirm whether this helps.

polypauldeep:
I tested the scenario described above using the current master branch and observed that the issue remains unresolved.

To reproduce the issue, I created a simple websocket server using the `websockets` library. 
1. This server sends a "wow" message to the clients every 500 ms once a connection is established. 
2. If the server receives a "hello" message from a client, it responds with a close message containing "wrong message."


```
import asyncio
import websockets

async def send_wow_periodically(websocket):
    try:
        while True:
            await websocket.send("wow")
            await asyncio.sleep(0.5)  # Wait for 500ms before sending the next message
    except websockets.exceptions.ConnectionClosed:
        print("Connection closed, stopping periodic messages.")

async def websocket_handler(websocket, path):
    task = asyncio.create_task(send_wow_periodically(websocket))

    try:
        async for message in websocket:
            if message == "hello":
[...]

Issue #942, opened 2023-08-12
ssl.SSLEOFError: EOF occurred in violation of protocol


gborrageiro-ld:
Hello,

I am getting an exception when subscribing to a high throughput websocket endpoint, `wss://fstream.binance.com/stream?streams=btcusdt@bookTicker/btcusdt@depth20@100ms/btcusdt@aggTrade/btcusdt@markPrice@1s/btcusdt@forceOrder`.
To be precise, it is the `btcusdt@bookTicker` that is very busy, and even if I switch off any computation completely and just run `websocket-client` and `rel`, eventually I run into the below exception.
```
File "...", line 179, in __init__
    rel.dispatch()
  File ".../python3.10/site-packages/rel/rel.py", line 228, in dispatch
    registrar.dispatch()
  File ".../python3.10/site-packages/rel/registrar.py", line 125, in dispatch
    if not self.loop():
  File ".../python3.10/site-packages/rel/registrar.py", line 134, in loop
    e = self.check_events()
  File ".../python3.10/site-packages/rel/registrar.py", line 285, in check_events
    self.callback('read', fd)
  File ".../python3.10/site-packages/rel/registrar.py", line 178, in callback
    self.events[etype][fd].callback()
  File ".../python3.10/site-packages/rel/listener.py", line 127, in callback
    if not self.cb(*self.args) and not self.persist and self.active:
  File ".../python3.10/site-packages/websocket/_app.py", line 437, in read
    op_code, frame = self.sock.recv_data_frame(True)
  File ".../python3.10/site-packages/websocket/_core.py", line 427, in recv_data_frame
    self.pong(frame.data)
  File ".../python3.10/site-packages/websocket/_core.py", line 351, in pong
    self.send(payload, ABNF.OPCODE_PONG)
  File ".../python3.10/site-packages/websocket/_core.py", line 283, in send
    return self.send_frame(frame)
  File ".../python3.10/site-packages/websocket/_core.py", line 311, in send_frame
    l = self._send(data)
  File ".../python3.10/site-packages/websocket/_core.py", line 525, in _send
    return send(self.sock, data)
  File ".../python3.10/site-packages/websocket/_socket.py", line 170, in send
    return _send()
  File ".../python3.10/site-packages/websocket/_socket.py", line 147, in _send
    return sock.send(data)
  File ".../python3.10/ssl.py", line 1206, in send
    return self._sslobj.write(data)
[...]

Issue #974, opened 2024-03-08
why on_close only was called once, so it causes the client to reconnect automatically only once, even I set reconnect parameter.


GooVincent:
Hi, everyone, I am trying to implement reconnecting ws server when close by the ws server. What I am using websocket-client version is 1.7.0. And I found on_close was just triggerd once when the ws server closed the socket.

Following is my demo code.

client.demo
``` client
def on_close():
       print('on_close is triggered')
        ws.run_forever(dispatcher=rel, reconnect=5)
        rel.signal(2, rel.abort)  # Keyboard Interrupt
        rel.dispatch()

def main()
      ws = websocket.WebSocketApp(
                    f"ws://{host}/ws",
                    on_open=on_open,
                    on_message=on_message,
                    on_close=on_close,
                    on_error=on_error,
                    ping_timeout=3,
                    ping_interval=5)
      
       ws.run_forever(dispatcher=rel, reconnect=5)
       rel.signal(2, rel.abort)  # Keyboard Interrupt
       rel.dispatch()
```

server.demo
``` server
async def test(ws):
    print(f'running test..')
    while True:
        current_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
        await asyncio.sleep(3)
        print(f"{current_time}")
        break


async def main():
    ip = "localhost"
    port = 8765
    async with websockets.serve(test, ip, port):
        print(f'ws server running on {ip}:{port}')
        await asyncio.Future()  
```

What I wanna is the client will keep reconnecting to the server, as I call run_forever() again in on_close, but it only happen once, why?

Please anyone could help me on this?

bubbleboy14:
Hey @GooVincent, a few notes:

1) only call rel.dispatch() once (in your example, main()) looks like the correct place)
2) no need to call rel.signal() - websocket-client takes care of that here: https://github.com/websocket-client/websocket-client/blob/a394b46c560cb78672f95bbb3836305e8511d1b9/websocket/_app.py#L142
3) ping_timeout and ping_interval kwargs should be passed to run_forever(), not the WebSocketApp constructor

If you provide complete (runnable via "python client.py") client and server examples, I can take a crack at your issue.

GooVincent:
Here is my complete client.py
```
import websocket
import rel
from functools import partial


class WsClient(object):
    def __init__(self) -> None:
        # websocket.enableTrace(True)
        host = "localhost:8765"

        self.ws = websocket.WebSocketApp(
[...]

PR #983, opened 2024-04-23
Asyncpong

bubbleboy14:
This PR consists of a few things, with the general intention of improving asynchronous writes and reconnect behavior.

TL/DR: fixes https://github.com/websocket-client/websocket-client/issues/976 and https://github.com/websocket-client/websocket-client/issues/942 and https://github.com/websocket-client/websocket-client/issues/974

background
----------
Running asynchronously (with rel dispatcher), I noticed a couple things.

Firstly, I was seeing an error (SSLEOFError) that occasionally occurred in the course of a write, in particular while sending a PONG (hence the name of this branch). I actually think I saw some ticket about this a while ago, but wasn't able to reproduce it at the time.

Secondly, I noticed that my on_reconnect callback wasn't being fired, and that reconnects weren't reliable - I had to handle some cases in an on_error handler.

reconnects
----------
The reconnect-related fix is in WrappedDispatcher: the problem was that reconnect() (by way of timeout()) wasn't passing True (as reconnecting kwarg) to setSock(); I modified reconnect() and timeout() to address this - problem solved.

async writes
------------
I figured this was happening due to my asynchronous usage of the socket - it was probably trying to write when there wasn't room in the write buffer or writes weren't available for some other reason. For reference, WebSocket.send_frame() (in the _core module, on line 340) is where an individual frame is sent.

The fix is to write asynchronously in custom dispatcher / async mode. I modified WebSocket._send() (line 562) to use the send() function on the dispatcher. DispatcherBase now has a send() function, which just calls _socket.send() (as WebSocket._send() did previously - so there is no change with the default sync dispatcher). WrappedDispatcher.send() uses dispatcher.buffwrite(), which is available in the latest version (0.4.9.17) of rel. The WebSocket constructor now takes a dispatcher kwarg, and saves it on self.

Back in the _app module, in WebSocketApp, create_dispatcher() takes a handleDisconnect kwarg, which is passed to WrappedDispatcher. Also in run_forever(): handleDisconnect(WebSocketConnectionClosedException, bool(reconnect)) lambda is passed (as handleDisconnect kwarg) to create_dispatcher() (which WrappedDispatcher.send() passes to buffwrite()); and setSock() passes the dispatcher to the WebSocket instance.

why the _dispatcher module?
--------------------------
I moved DispatcherBase, Dispatcher, SSLDispatcher, and WrappedDispatcher to the new _dispatcher module so that the _core module could non-circularly import DispatcherBase and WrappedDispatcher for the dispatcher kwarg type hint in the WebSocket constructor.

status
------
This is still something of a work in progress. I've been testing this branch, and for my use case it's performing better than the current master branch, but further testing would be a good idea.

Please test with your use case and LMK how it goes!

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -501,7 +501,7 @@ class WebSocketApp:
         custom_dispatcher = bool(dispatcher)
         dispatcher = self.create_dispatcher(
             ping_timeout, dispatcher, parse_url(self.url)[3],
-            lambda : handleDisconnect(WebSocketConnectionClosedException, bool(reconnect))
+            lambda e : handleDisconnect(WebSocketConnectionClosedException(e), bool(reconnect))
         )
 
         try:
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -468,7 +468,7 @@ class WebSocketApp:
                     )
                 ):
                     raise WebSocketTimeoutException("ping/pong timed out")
-            return TruehandleDisco
+            return True
 
         def closed(
             e: Union[
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -481,7 +481,7 @@ class WebSocketApp:
                 fire_cont_frame=self.on_cont_message is not None,
                 skip_utf8_validation=skip_utf8_validation,
                 enable_multithread=True,
-                dispatcher: dispatcher,
+                dispatcher=dispatcher,
             )
 
             self.sock.settimeout(getdefaulttimeout())
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -429,7 +429,7 @@ class WebSocketApp:
                     raise e
 
             if op_code == ABNF.OPCODE_CLOSE:
-                return teardown(frame)
+                return closed(frame)
             elif op_code == ABNF.OPCODE_PING:
                 self._callback(self.on_ping, frame.data)
             elif op_code == ABNF.OPCODE_PONG:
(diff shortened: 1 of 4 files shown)


ErnestoLara (approved):

bubbleboy14:
> any updates on this? can we get someone to fix the test errors, review and merge?

@jack-rep does this branch improve async performance for you?

bubbleboy14:
@almostDemoPy has confirmed that this PR fixes  https://github.com/websocket-client/websocket-client/issues/942#issuecomment-2301187309, and AFAICT it fixes https://github.com/websocket-client/websocket-client/issues/976 and https://github.com/websocket-client/websocket-client/issues/974 as well.

Has anyone else tested? @engn33r does this branch seem to work with your use cases?

bubbleboy14:
> I haven't looked at this yet even though it's been here a while...
> 
> @bubbleboy14 can you fix the lint errors from flake8 and black so the CI at least partially passes? After that, I'll test out the changes to confirm it's working for me as expected and we can merge this

Good call @engn33r! I fixed the flake8 and black lint errors, LMK if there's anything else I should adjust.

thomaska:
If I may ask is there a plan or a date of when this fix would be released in a new version of the library?
We've been hitting the "EOF occured in violation" error quite a lot, so we would be eager to get this fix! 
cc @bubbleboy14

engn33r:
@thomaska v1.9.0 will be released in the next few days

PR #998, opened 2024-08-30
Include pytest in test extra

carlwgeorge:
Fedora uses the test extra to automatically generate build-time dependencies, so including pytest in this extra allows removal of an explicit BuildRequires in the RPM spec file.


engn33r:
If it helps, we'll add it in. Thanks for the PR.

PR #999, opened 2024-09-04
Adding logic to allow the client to decide if the websocket frame mask should be used when sending messages.

QuinnDamerell:
After debugging perf issues on lower-end devices like a Pi4 and Pi3, I found that the WebSocket lib was performing a masking operation over the entire buffer sent, as the websocket standard requires. Unfortunately, that masking operation is quite costly; I measured that it doubled the amount of time required to send a ~10kb buffer. 

I looked into why the masking is done, and it was added due to the [security issue discussed in this paper](https://web.archive.org/web/20180127212141/http://w2spconf.com/2011/papers/websocket.pdf). The problem is that the paper was published in 2011 and deals with unencrypted web sockets. Since almost all internet traffic WebSockets now use SSL, there have been a [few attempts to remove the masking from the protocol](https://www.ietf.org/archive/id/draft-damjanovic-websockets-nomasking-00.html) since it adds overhead and isn't needed anymore.

I added a simple optional flag to the send call that defaults to masking enabled (as the standard requires), but it allows the client to disable the mask if desired. As I said above, if the client knows the websocket is opened via SSL or it's only being used on a local LAN, the mask is not needed. As I also said, skipping the mask resulted in dropping the send time by half and reducing the CPU percentage required by 10% in my application. 

From my testing, most modern web servers will accept client messages without masking, so this seems to be a helpful option for lower-end IOT devices.

If there are any questions, I would love to discuss them!

### DIFF
--- a/websocket/_abnf.py
+++ b/websocket/_abnf.py
@@ -205,7 +205,7 @@ class ABNF:
         return f"fin={self.fin} opcode={self.opcode} data={self.data}"
 
     @staticmethod
-    def create_frame(data: Union[bytes, str], opcode: int, fin: int = 1) -> "ABNF":
+    def create_frame(data: Union[bytes, str], opcode: int, fin: int = 1, use_frame_mask: bool = True) -> "ABNF":
         """
         Create frame to send text, binary and other data.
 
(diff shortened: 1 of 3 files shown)


engn33r:
I'll look again soon to collect any final comments before merging, but I wanted to say this looks good, thanks for the PR :+1:

PR #1001, opened 2024-09-13
Update _app.py

239144498:
Adding functionality to ping doesn't work well if it can only be a string type, when adding a function produces better results, for example: old "ping" -> run_forever(ping_payload="ping")
new "ping:1726197303" -> run_forever(ping_payload=lambda : f "ping:{int(time.time()}")) I think the modification will make the code work better

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -335,7 +335,7 @@ class WebSocketApp:
                 self.last_ping_tm = time.time()
                 try:
                     _logging.debug("Sending ping")
-                    self.sock.ping(self.ping_payload)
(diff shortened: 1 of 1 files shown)


UUSim:
Why not use `callable` instead of `inspect.isfunction`?

Note that you can already have a dynamic ping body using this trick:
```python
class PingStr(str):
    """ Class to generate dynamic ping payload message. """

    def encode(self, encoding='utf-8', errors='strict'):
        # insert custom body here
        return f"ping:{int(time.time()}"
```

Then call `run_forever` with `ping_payload=PingStr("")`.

PR #1006, opened 2024-10-30
Fix type of on_ parameters

Shutgun:
I think, the typing of the "on_" parameters of WebsocketApp is not correct. According to the doc strings and what I see in the debugger, the first parameter is typically the class object itself.


engn33r:
Thank you for the PR. The type hints in this library are still a work in progress, there are likely to be other similar changes needed.

PR #1015, opened 2025-01-14
Run forever socket close fix

justmobilize:
In cases like `handleDisconnect`, the `on_error` callback is called first. If the `WebSocketApp` is being used in a multithreaded application, we nee to make sure that `self.sock` is cleared before calling close, otherwise logic built around the sock being set can cause issues - specifically calling `run_forever` again, since is checks if `self.sock` is set.

This was found when using the 3rd party library [pysher](https://github.com/deepbrook/Pysher) that uses the websocket-client

As for example code, it isn't particularly easy. but if it's needed I'm sure I can put something together.


engn33r:
I am getting some AI assistance to clean up this codebase. It found 4 bugs and this fixes one of the identified ones.

```
THREAD SAFETY RACE CONDITION (_app.py:350-365)
    - Impact: Potential AttributeError when accessing sock after it's set to None
    - Root Cause: sock access outside lock protection in teardown method
```

Here's the test that was generated to demonstrate the issue:

```python
import threading
import time
import unittest
from unittest.mock import Mock, patch, MagicMock, call
import sys

import websocket
from websocket import WebSocketApp
from websocket._exceptions import (
    WebSocketConnectionClosedException, 
    WebSocketException,
    WebSocketTimeoutException
)
from websocket._abnf import ABNF


class WebSocketAppCriticalBugsTest(unittest.TestCase):
    """Test critical bugs and coverage gaps in WebSocketApp"""

    def test_bug_thread_safety_teardown_race_condition(self):
        """
        BUG: Test race condition in teardown method accessing sock outside lock
        
        Bug location: _app.py:350-365 - sock access outside lock
        """
        app = WebSocketApp("ws://example.com/ws")
        app.sock = Mock()
        
        # Simulate concurrent teardown calls
        results = []
        
        def concurrent_teardown():
            try:
                app._run_forever_teardown(None)  # This calls the internal teardown
                results.append("success")
            except AttributeError as e:
                if "'NoneType'" in str(e):
                    results.append("race_condition")
                else:
                    results.append(f"other_error: {e}")
        
        # Create teardown function that mimics the internal one
        def teardown_with_race_condition():
            with app.has_done_teardown_lock:
                if app.has_done_teardown:
                    return
                app.has_done_teardown = True
            
            # Simulate race condition - one thread sets sock to None
            # while another tries to access it
            if app.sock:
                # Thread 1 might set sock to None here
                app.sock = None
            
            # Thread 2 tries to access sock here (outside the lock!)
            if app.sock:  # This check can pass...
                app.sock.close()  # But this can fail if sock becomes None
        
        app._run_forever_teardown = teardown_with_race_condition
        
        # Start multiple threads to trigger race condition
        threads = []
        for _ in range(5):
            t = threading.Thread(target=concurrent_teardown)
            threads.append(t)
            t.start()
        
        for t in threads:
            t.join()
        
        # At least one thread should succeed
        self.assertIn("success", results)

if __name__ == "__main__":
    unittest.main()
```

engn33r:
Thanks for the helpful PR!

Issue #970, opened 2024-02-18
on_error callback stop being called after first reconnection


jeanbmar:
After a failure, this line will set the WebSocketApp instance in reconnecting mode forever:

https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L613

Indeed, `setSock` will only be called as reconnecting. As a result any further error will be swallowed here:

https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L582

I don't think we even want to swallow the reconnection-specific errors. Should the condition be completely removed?

Ankit9608:
Same doubt after first connection drops it silently reconnect without calling on_error(), i am using run_forever(reconnect=5) what is solution on this. one more while reconnecting does it completly create new tcp connection and close previous one or reconnect with existing basically, while reconnecting on_close() is called or not?

PR #1016, opened 2025-01-31
Always call on_error

KoenVda88:
on_error should be called even when we are in reconnected state.

If not the on_error will be called when we lose the connection the first time but not the second time


engn33r:
If I understand correctly, the intent here is to resolve #970. I need to check if this is the best solution or if another approach is better.

KoenVda88:
FYI, I installed V1.9.1 and I still have this issue. The first time I plug and unplug the cable to my device the reconnect works. But the second time it just stays in connected state. 
This change solves it.

engn33r:
I agree that this PR fixes a real issue, but I will take some more time before merging to make sure it does not introduce any unexpected edge cases. You can expect to see this included in 1.10.0

PR #1033, opened 2025-09-11
Use ipaddress module for no_proxy parsing

jplitza:
This enables support for IPv6 CIDRs in the no_proxy option

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -478,7 +478,7 @@ class WebSocketApp:
                 SystemExit,
                 Exception,
                 str,
-            ] = "closed unexpectedly"
+            ] = "closed unexpectedly",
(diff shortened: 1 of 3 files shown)


engn33r:
I cannot add my changes to this to fix failing CI @jplitza because I don't have proper permissions. But you need to apply black formatting and fix the AssertionError in test_connect to make CI pass.

If this is not changed in 2 weeks, I will make a duplicate PR with a similar edit, with the CI fixes.

jplitza:
That should do it. Those black findings were both outside what I changed, so I hoped rebasing would fix it, but it didn't.

engn33r:
Thanks for the PR, and the quick fix!

PR #1039, opened 2025-10-16
Adding Minor Perf Improvements.

QuinnDamerell:
Hey!

I pulled over a few changes from my fork that are easy to merge in, and I thought you might want them, too.

I also added logic for zero buffer copy [sends](https://github.com/OctoEverywhere/octowebsocket-client/commit/ab32c092fca02c970d7dbdcf0cc23257ecb2750b) and [receives](https://github.com/OctoEverywhere/octowebsocket-client/commit/9fdf16ec6fc5f1f820655d721710a66c1cdd5a20) in my fork. If you're interested in those, I can pull them over, too. They involve too many changes.

### DIFF
--- a/websocket/_dispatcher.py
+++ b/websocket/_dispatcher.py
@@ -74,12 +74,11 @@ class Dispatcher(DispatcherBase):
             return
         with selectors.DefaultSelector() as sel:
             sel.register(self.app.sock.sock, selectors.EVENT_READ)
-            try:
-                while self.app.keep_running:
-                    if sel.select(self.ping_timeout):
-                        if not read_callback():
-                            break
-                    check_callback()
+            while self.app.keep_running:
+                if sel.select(self.ping_timeout):
+                    if not read_callback():
+                        break
+                check_callback()
 
 
 class SSLDispatcher(DispatcherBase):
@@ -98,12 +97,11 @@ class SSLDispatcher(DispatcherBase):
         sock = self.app.sock.sock
         with selectors.DefaultSelector() as sel:
             sel.register(sock, selectors.EVENT_READ)
-            try:
-                while self.app.keep_running:
-                    if self.select(sock, sel):
-                        if not read_callback():
-                            break
-                    check_callback()
+            while self.app.keep_running:
+                if self.select(sock, sel):
+                    if not read_callback():
+                        break
+                check_callback()
 
     def select(self, sock, sel: selectors.DefaultSelector):
         if self.app.sock is None:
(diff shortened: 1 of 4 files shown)


websocket/_http.py
```diff
@@ -88,6 +88,8 @@ def _start_proxied_socket(url: str, options, proxy) -> tuple:
 
     hostname, port, resource, is_secure = parse_url(url)
 
+    proxy_type = None
+    rdns = False
```
engn33r, line 92: This line is already found inside the first if statement and in the sock5 elif. Should it be removed from those places to avoid duplication?

websocket/_http.py
```diff
@@ -88,6 +88,8 @@ def _start_proxied_socket(url: str, options, proxy) -> tuple:
 
     hostname, port, resource, is_secure = parse_url(url)
 
+    proxy_type = None
```
engn33r, line 91: Is this line necessary? If so, I don't understand why it is needed

websocket/_dispatcher.py
```diff
@@ -98,16 +95,13 @@ def read(
...
-        sel = selectors.DefaultSelector()
-        sel.register(sock, selectors.EVENT_READ)
-        try:
+        with selectors.DefaultSelector() as sel:
+            sel.register(sock, selectors.EVENT_READ)
             while self.app.keep_running:
                 if self.select(sock, sel):
                     if not read_callback():
                         break
                 check_callback()
```
engn33r, line 104: There was a `sel.close()` in the finally block in the dispatchers that is now removed. Can you explain why those can be removed?

websocket/tests/test_dispatcher.py
```diff
@@ -76,6 +78,13 @@ def buffwrite(self, sock, data, send_func, disconnect_handler):
     def timeout(self, seconds, callback, *args):
         self.timeout_calls.append((seconds, callback, args))
 
+    def __enter__(self):
```
engn33r, line 81: Does \_\_enter\_\_ and \_\_exit\_\_ get called anywhere? I don't understand the purpose of this addition

engn33r (commented):
Thanks for the PR! I might be missing obvious things, but a bit more explanation of some of these changes will help me understand what they are supposed to do. afaik no features are added and I'm unclear on how some of these edits are performance improvements

websocket/_http.py
QuinnDamerell: I'm not actually sure in PY, but in most languages, you would want to declare the var outside of the scope, since it's used outside of the scope. Consider, for example, if someone added an ifelse case and didn't realize they needed to set it, then after that case, it would be undeclared.

PR #1040, opened 2025-11-10
Fix #970, call on_error even after first error

### LINKED ISSUE #970: on_error callback stop being called after first reconnection

engn33r:


engn33r:
Closing in favor of PR #1016 because this PR attempts to resolve the same issue that 1016 does better

Issue #1055, opened 2026-08-31
WebSocketApp.close() raises AttributeError: 'NoneType' object has no attribute 'close_frame' (regression in 1.9.1)


r-madsen-qdi:
This is a small example reproducing the issue [repro_close_frame.py](https://github.com/user-attachments/files/31643824/repro_close_frame.py)

close() reads self.sock twice — once to guard, once after self.sock.close(**kwargs) has returned:

# websocket/_app.py:207
self.keep_running = False
if self.sock:
    self.sock.close(**kwargs)
    if self.sock.close_frame is not None:   # ← line 211
        self.last_close_frame = self.sock.close_frame
    self.sock = None

WebSocket.close() (_core.py:538) sets connected = False, sends the CLOSE frame, then blocks in a recv_frame() loop for up to timeout=3 seconds awaiting the peer's echo. During that
window the run_forever() thread's own recv fails and reaches teardown(), which sets self.sock = None (_app.py:387). Line 211 then dereferences None.

teardown() already handles this correctly, and its comment states the reason:

# _app.py:385
current_sock = self.sock
self.sock = None      # cleared BEFORE close, for multithreaded safety
current_sock.close()

close() didn't get the same treatment.

Introduced by 10d9159af ("Improve close frame handling to fix #1043 and add tests", #1044, 2026-01-27), shipped in 1.9.1 (2026-08-29). The test added alongside it,
test_client_initiated_close_captures_peer_close_frame, drives close() with self.sock = mock.Mock(); a mock's close() doesn't null app.sock, so the single-threaded test passes and the
concurrent path is uncovered.

Repro (attached): 21/25 on 1.9.1, 0/25 on 1.9.0, Python 3.12.3.

Suggested fix — capture into a local, matching teardown():

self.keep_running = False
sock = self.sock
if sock:

engn33r:
Thank you for reporting this. This looks like a legitimate bug, I will aim to resolve this very soon

engn33r:
Resolved by #1057 and should be fixed in release 1.9.2

PR #1057, opened 2026-08-31
Fix race in WebSocketApp.close() when run_forever thread clears sock

### LINKED ISSUE #1056: version 1.9.1 now crashes with AttributeError in WebSocketApp.close()
In version 1.9.0 the `WebSocketApp.close()` method looked like this
```
    def close(self, **kwargs) -> None:
        """
        Close websocket connection.
        """
        self.keep_running = False
        if self.sock:
            self.sock.close(**kwargs)
            self.sock = None
```

In version 1.9.1 this method now looks like this
```
    def close(self, **kwargs: Any) -> None:
        """
        Close websocket connection.
        """
        self.keep_running = False
        if self.sock:
            self.sock.close(**kwargs)
            # Capture the peer's close frame before clearing socket reference
            if self.sock.close_frame is not None:
                self.last_close_frame = self.sock.close_frame
            self.sock = None
```

The `if self.sock.close_frame is not None` is now crashing with an `AttributeError`. I'm a developer for TrueNAS and we have used this library to underpin our Websocket client and our CI/CD has now started to crash here at this line. Our client runs `run_forever()` in a background thread and when we call our own `Client.close()` it calls `WebSocketApp.close()` from the main thread. While `self.sock.close(**kwargs)` is doing the close handshake, the background thread notices the connection closing and runs `teardown()`.  When the main thread comes back and evaluates `self.sock.close_frame`, the `self.sock` attribute is now `None`

For posterity, here is what our traceback looks like that we're seeing in our CI/CD.
```
Traceback (most recent call last):
  File "/var/lib/jenkins/workspace/master/clean/jenkins/scripts/builder_orchestrator.py", line 2362, in <module>
    main()
    ~~~~^^
  File "/var/lib/jenkins/workspace/master/clean/jenkins/scripts/builder_orchestrator.py", line 2358, in main
    client.close()
    ~~~~~~~~~~~~^^
  File "/var/lib/jenkins/workspace/master/clean/.venv/lib/python3.13/site-packages/truenas_api_client/__init__.py", line 1008, in close
    self._ws.close()
    ~~~~~~~~~~~~~~^^
  File "/var/lib/jenkins/workspace/master/clean/.venv/lib/python3.13/site-packages/truenas_api_client/__init__.py", line 220, in close
    self.app.close()
    ~~~~~~~~~~~~~~^^
  File "/var/lib/jenkins/workspace/master/clean/.venv/lib/python3.13/site-packages/websocket/_app.py", line 211, in close
    if self.sock.close_frame is not None:
       ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'close_frame'

yocalebo:
Version 1.9.1 introduced a crash when WebSocketApp.close() is called from a different thread than the one running run_forever(). Commit 10d9159 (PR #1044) changed close() so that it reads self.sock.close_frame after calling self.sock.close(). While self.sock.close() is waiting for the close handshake to finish, the run_forever() thread notices the connection is closing, runs teardown() and sets self.sock to None. When close() then reads self.sock.close_frame it raises AttributeError because self.sock is gone. Versions up to 1.9.0 never touched self.sock again after closing it, so this crash could not happen there. The fix is for close() to keep a local reference to the socket and use that for the whole method, so it never reads self.sock after another thread has cleared it.

Fixes #1056

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -205,11 +205,12 @@ class WebSocketApp:
         Close websocket connection.
         """
         self.keep_running = False
-        if self.sock:
-            self.sock.close(**kwargs)
+        sock = self.sock
+        if sock:
+            sock.close(**kwargs)
             # Capture the peer's close frame before clearing socket reference
-            if self.sock.close_frame is not None:
-                self.last_close_frame = self.sock.close_frame
+            if sock.close_frame is not None:
+                self.last_close_frame = sock.close_frame
             self.sock = None
 
     def _start_ping_thread(self) -> None:
--- a/websocket/tests/test_app.py
+++ b/websocket/tests/test_app.py
@@ -784,6 +784,46 @@ class WebSocketAppUnitTests(unittest.TestCase):
 
         self.assertEqual(close_results, [(1000, "peer-goodbye")])
 
+    def test_close_captures_frame_when_teardown_clears_socket(self):
+        """close() retains its socket while concurrent teardown clears app.sock."""
+        app = self._build_app()
+        peer_close_frame = ABNF.create_frame(
+            struct.pack("!H", 1000) + b"peer-goodbye", ABNF.OPCODE_CLOSE
+        )
+        close_started = threading.Event()
+        allow_close_to_return = threading.Event()
+        errors = []
+
+        class FakeWebSocket:
+            close_frame = None
+
+            def close(self, **kwargs):
+                self.close_frame = peer_close_frame
+                close_started.set()
+                if not allow_close_to_return.wait(timeout=1):
+                    raise TimeoutError("test did not release FakeWebSocket.close()")
+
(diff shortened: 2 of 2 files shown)


yocalebo:
Interestingly enough, after looking at other open PRs on this repo I've noticed this PR that has been rotting with no input/review https://github.com/websocket-client/websocket-client/pull/1046. It pointed out the same regression and tried to fix it using a lock.

engn33r:
This PR fixes #1056 and #1055 (they report the same issue). As mentioned above, the issue was introduced in #1044, so only release 1.9.1 is affected. Release 1.9.2 will include this PR and resolve the issue.

The original PR resolves the issue, but I added a regression test to help prevent this issue from reappearing in the future.

PR #1060, opened 2026-09-04
fix auto gc websocketapp 

lubupang:
```

import threading,json,time
import rel
from websocket import WebSocketApp
import websocket
websocket.enableTrace(True)
wsurl='ws://127.0.0.1:57915/devtools/browser/590396f0-f89d-4c23-b387-d99cdca223e6' #replace as yours
class A:
    ...

a=A()
def _on_message(ws,data):
    print(data)



a._websocket=WebSocketApp(wsurl,on_message=_on_message)

print(rf'''{id(a._websocket)=}''')  # the old one will never auto gc

t=threading.Thread(target=a._websocket.run_forever,kwargs={'reconnect':1,'ping_timeout':10.0},daemon=True)
t.start()
while not a._websocket.ready():
    time.sleep(0.1)
a._websocket.send(json.dumps({"id":1,"method":"Browser.getVersion","params":{}}))
time.sleep(10) #make sure no error when recv
a._websocket=WebSocketApp(wsurl,on_message=_on_message)  # new instance 
print(rf'''{id(a._websocket)=}''')

_websocket=WebSocketApp(wsurl,on_message=_on_message) 
print(rf'''{id(_websocket)=}''')
_websocket.run_forever(dispatcher=rel, reconnect=5)  # Set dispatcher to automatic reconnection, 5 second reconnect delay if connection closed unexpectedly
rel.signal(2, rel.abort)  # Keyboard Interrupt
rel.dispatch()
```

`WebSocketApp` cannot be automatically garbage-collected.
If the application no longer has an outer reference to the `WebSocketApp`, there is no way to access the app and call `app.close()`. so the `app.keep_running` will always `True`.
dispatcher will try reconnect until ```(KeyboardInterrupt, SystemExit)```  
top code shows the old one can not auto gc 1.  but the new can gc the first app. the 2. and 3. do not need gc.

### DIFF
--- a/websocket/_app.py
+++ b/websocket/_app.py
@@ -2,6 +2,8 @@ import inspect
 import socket
 import threading
 import time
+import gc
+import types
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 from ._logging import debug, error, info, warning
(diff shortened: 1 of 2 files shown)


lubupang:
If there are no outer references to the `WebSocketApp`, the application can no longer call methods such as `app.send(xxx)`. Therefore, I think the `WebSocketApp` is useless at this point and should be able to be garbage-collected.

Someone could still obtain a reference through `app.sock` and send something directly, but this seems like an edge case that should be avoided as much as possible.

lubupang:
open app `def __del__`   will find that

lubupang:
Solve the problem with the least amount of code. ha ha ha ha ha ha ha ha...

lubupang:
In the long run, refactoring `run_forever` would be the better approach, but the hidden costs and risks are simply too high for this particular issue.

PR #1063, opened 2026-09-19
Ignore invalid cookies in handshake responses

### LINKED ISSUE #1061: Unparseable Set-Cookie aborts an already-successful handshake
The `handshake_response.__init__` calls `CookieJar.add(headers.get("set-cookie"))` with no error handling. `SimpleCookieJar.add` calls `http.cookies.SimpleCookie(...)`, which the stdlib documents as raising `CookieError` on data deemed invalid: "_you should always prepare for invalid data and catch CookieError on parsing_". So a malformed cookie from a server kills a connection that already returned `101 Switching Protocols`. Perhaps, the exception should be caught ignored (after logging it).

It may also be good to allow the caller to specify, that cookies should be completely ignored for a particular connection.

jakezwang:
Fixes #1061.

A malformed `Set-Cookie` header can raise `CookieError` after a successful WebSocket upgrade. Ignore that header and log a warning without its contents, so cookie parsing does not abort the connection or a redirect.

Regression tests cover the upgrade, existing-cookie preservation, warning output, and redirect response. Valid cookie handling is unchanged.

Validation: full tests on Python 3.10 with the local echo server (207 passed, 16 skipped) and Python 3.14 (197 passed, 26 skipped); focused tests on 3.11–3.13; three real loopback handshakes; Black, doctests, build, and Twine checks. Internet-dependent unit tests were skipped.


engn33r:
lgtm, thanks for the commit

